PR #66, opened 2020-01-17
Fix #65 Don’t break idempotency if vid is declared as string

### LINKED ISSUE #65: Idempotency is broken if vid is declared as a string
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
When a vid is declared with a string as its vid, the vlan is correctly created in netbox but idempotency is broken.
before will be int and after will be string so changed will always be true.

This is also the case for other modules.

##### STEPS TO REPRODUCE

```yaml
- name: "netbox add vlans"
  netbox_community.ansible_modules.netbox_vlan:
    netbox_token: "{{ netbox['token'] }}"
    netbox_url: "{{ netbox['url'] }}"
    validate_certs: "no"
    data:
      status: "Active"
      vid: "{{ item['vlanid'] }}"
      name: "{{ item['name'] }}"
      site: "{{ item['location'] }}"
  connection: local
  delegate_to: localhost
  loop: "{{ subnet|flatten(levels=1) }}"
```

```yaml
 - vlanid: "42"
  name: "vlaname"
  location: "xxx"
```

##### EXPECTED RESULTS
```
    "ansible_loop_var": "item",
    "changed": false,
    "diff": {
        "after": {},
        "before": {}
    },
```

##### ACTUAL RESULTS
```
    "ansible_loop_var": "item",
    "changed": true,
    "diff": {
        "after": {
            "vid": "42"
        },
        "before": {
            "vid": 42
        }
    },
```

FurryJulie:
In modules_utils/netbox_ipam.py _build_diff():
When an already present vlan with the same data is pushed, there’s a
diff between vid: before is a int and after is a string.
Changed = True if diff is instanced so every values in before
and after are converted to string in order to compare them. If they are equals,
diff = None.

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -455,7 +455,7 @@ class NetboxModule(object):
             after_str[k] = str(v)
 
         if before_str == after_str:
-            return None
+            return {"before": {}, "after": {}}
         else:
             return {"before": before, "after": after}
 
@@ -746,7 +746,8 @@ class NetboxModule(object):
                 self._handle_errors(
                     msg="Request failed, couldn't update device: %s" % name
                 )
-            if diff:
+            if diff and (len(diff['before'].keys()) > 0 or\
+                         len(diff['after'].keys()) > 0):
                 self.result["msg"] = "%s %s updated" % (endpoint_name, name)
                 self.result["changed"] = True
                 self.result["diff"] = diff
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -445,7 +445,19 @@ class NetboxModule(object):
 
     def _build_diff(self, before=None, after=None):
         """Builds diff of before and after changes"""
-        return {"before": before, "after": after}
+        before_str = {}
+        after_str = {}
+
+        for k, v in before.items():
+            before_str[k] = str(v)
+
+        for k, v in after.items():
+            after_str[k] = str(v)
+
+        if before_str == after_str:
+            return None
+        else:
+            return {"before": before, "after": after}
 
     def _convert_identical_keys(self, data):
         """


FragmentedPacket:
@FurryJulie Thanks for the work on this. I think the way I'd want to handle this is using the argument spec which I'm looking into now. I think I'll have to create the argument spec for each module and then add in some special processing to not have it send any other data than what was specified since the argument spec provides defaults (None) which would null out options on the specific objects.


Just an idea I'm exploring atm:

```python
    argument_spec = dict(
        netbox_url=dict(type="str", required=True),
        netbox_token=dict(type="str", required=True, no_log=True),
        data=dict(
            type="dict",
            required=True,
            options=dict(
                site=dict(required=False, type="raw"),
                vlan_group=dict(required=False, type="raw"),
                vid=dict(required=True, type="int"),
                name=dict(required=True, type="str"),
                tenant=dict(required=False, type="raw"),
                status=dict(
                    required=False, choices=["Active", "Reserved", "Deprecated"],
                ),
                vlan_role=dict(required=False, type="raw"),
                description=dict(required=False, type="str"),
                tags=dict(required=False, type="list"),
                custom_fields=dict(required=False, type="dict"),
            ),
        ),
        state=dict(required=False, default="present", choices=["present", "absent"]),
        validate_certs=dict(type="bool", default=True),
    )
```

Here is the playbook:
```yaml
---
- name: "Testing arg specs"
  hosts: localhost
  connection: local
  gather_facts: no

  tasks:
    - name: "Test idempotency"
      netbox_community.ansible_modules.netbox_vlan:
        netbox_url: "http://netbox-demo.org:32768"
        netbox_token: "token"
        data:
          name: "Test VLAN"
          vid: "400"
```

Here is the debugged output:
```
ok: [localhost] => changed=false
  args:
    data:
      custom_fields: null
      description: null
      name: Test VLAN
      site: null
      status: null
      tags: null
      tenant: null
      vid: 400
      vlan_group: null
      vlan_role: null
    netbox_token: VALUE_SPECIFIED_IN_NO_LOG_PARAMETER
    netbox_url: http://netbox-demo.org:32768
    state: present
    validate_certs: true
  invocation:
    module_args:
      data:
        custom_fields: null
        description: null
        name: Test VLAN
        site: null
        status: null
        tags: null
        tenant: null
        vid: 400
        vlan_group: null
        vlan_role: null
      netbox_token: VALUE_SPECIFIED_IN_NO_LOG_PARAMETER
      netbox_url: http://netbox-demo.org:32768
      state: present
      validate_certs: true
```

You can see that the argument spec says the **vid** should be **int** type and does the conversion for us. I think that may be a better long term solution.

FragmentedPacket:
@FurryJulie I have the `arg-specs` branch. Could you test and let me know if it fixes your idempotency issue for netbox_vlan?

FurryJulie:
Thanks for the reply!

Yes it seems way cleaner than what I proposed.

The idempotency of the branch `arg-specs` works nicely in my case with the modules ansible_vlans, thanks :)

It seems there’s the same issue with ansible_prefix so I think it can be fixed the same way:

```
    "changed": true,
    "diff": {
        "after": {
            "family": "4"
        },
        "before": {
            "family": 4
        }
    }
```

FragmentedPacket:
Thanks for checking the branch. I will be going through each module to implement it, but if you find any other modules such as the `netbox_prefix` above, please let me know on here just to make sure I get it.

I will hopefully get this finished by next week and have it pushed into the `develop` branch and then hopefully release a new version ASAP.

FragmentedPacket:
@FurryJulie Thanks for your work on this. I just merged PR #68 to address this. Hopefully this covers most everything, but if you see anything else, please open another issue and I can look into it.

Issue #58, opened 2020-01-13
Feature: Add Netbox Services


loganbest:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Feature could be used to configure monitoring systems by addition of services defined in Netbox in Ansible hostvars. Netbox allows arbitrary structured data to be assigned to servers via custom fields, tags, configuration contexts and services. 

Original Related Ansible Core Issues:
https://github.com/ansible/ansible/issues/47918
https://github.com/ansible/ansible/pull/47919

@ollybee we've pulled the Inventory plugin out of core and into this collection. Could you please submit a new PR to this issue with your changes?

FragmentedPacket:
@jhanssenfavaro Can you check the PR and see if it meets your requirements?

PR #70, opened 2020-01-29
Netbox service

TawR1024:
Hope it will help close an issue #58


tests/integration/integration-tests.yml
```diff
@@ -4079,6 +4079,54 @@
...
           - test_five['msg'] == "circuit_termination test_circuit_a deleted"
+          -
+
+##
+##
+### NETBOX_SERVICE
+##
+##
+
+    - name: "1 - Device with required information needs to add new service"
```
FragmentedPacket: Can you add two more tests? One to test idempotence, just try create the first one, but then create an identical test and make sure it reports that it already exists. And the other one is updating the service, make sure that is tested.

You can also just use an existing device that gets created within `tests/integration/netbox-deploy.py` for this as well.

plugins/module_utils/netbox_utils.py
```diff
@@ -306,10 +307,12 @@
     "nat_inside": set(["vrf", "address"]),
     "parent_region": set(["slug"]),
     "platform": set(["slug"]),
+    "port": set(["name"]),
```
FragmentedPacket: I don't think we need **port** or **protocol** here as they won't have to be searched.

plugins/module_utils/netbox_utils.py
```diff
@@ -392,6 +395,7 @@
     "vlan_groups",
 }
 
+
 NETBOX_ARG_SPEC = dict(
```
FragmentedPacket: Can you add to the **REQUIRED_ID_FIND** for `protocol` and status and then create the constant to convert the status to an integer?

This eventually won't be needed, but at least for now.

plugins/modules/netbox_service.py
```diff
@@ -0,0 +1,181 @@
...
+DOCUMENTATION = r"""
+---
+module: netbox_service
+short_description: Creates or removes service from Netbox
+description:
+  - Creates or removes service from Netbox
+notes:
+  - This should be ran with connection C(local) and hosts C(localhost)
+author:
+  - Mikhail Yohman (@FragmentedPacket)
```
FragmentedPacket: You can remove me as an author since it's all you!

plugins/modules/netbox_service.py
```diff
@@ -0,0 +1,181 @@
...
+  - Kulakov Ilya (@TawR1024)
+requirements:
+  - pynetbox
+version_added: '2.8'
```
FragmentedPacket: Change version added to 0.1.5 since that is what it will be released under for collections.

plugins/modules/netbox_service.py
```diff
@@ -0,0 +1,181 @@
...
+    description:
+      - Defines the service configuration
+    suboptions:
+      device:
+        description:
+          - Specifies on which device the service is running
+        type: raw
+      virtual_machine:
+        description:
+        type: str
```
FragmentedPacket: Type for **virtual_machine** needs to be **raw**

plugins/modules/netbox_service.py
```diff
@@ -0,0 +1,181 @@
...
+      port:
+        description:
+          - Specifies which port used by service
+        type: int
+      protocol:
+        description:
+          - Specifies which protocol used by service
+        type: int
+        choises:
```
FragmentedPacket: Fix the typo: choises -> choices

And then just have the choices as the words and not integers since we do that in the background, even those they are valid at the moment.

plugins/modules/netbox_service.py
```diff
@@ -0,0 +1,181 @@
...
+            - 17 UDP
+      ipaddress:
+        description:
+          - VRF that prefix is associated with
+        type: str
+      description:
+        description:
+          - Service description
+        type: str
+      tags:
```
FragmentedPacket: It doesn't actually have tags available so you can remove that.

plugins/modules/netbox_service.py
```diff
@@ -0,0 +1,181 @@
...
+        description:
+          - Must exist in Netbox and in key/value format
+        type: dict
+    required: true
+  state:
+    description:
+      - Use C(present) or C(absent) for adding or removing.
+    choices: [ absent, present ]
+    default: present
+  first_available:
```
FragmentedPacket: You can remove first_available as well

plugins/modules/netbox_service.py
```diff
@@ -0,0 +1,181 @@
...
+  tasks:
+    - name: Delete service
+      netbox_service:
+        netbox_url: netbox_url  
+        netbox_token: netbox_token
+        data:
+          device: Test666
+          name: node-exporter
+          port: 9100
+          protocol: 6
```
FragmentedPacket: Update the protocol to words.

plugins/modules/netbox_service.py
```diff
@@ -0,0 +1,181 @@
...
+        dict(
+            data=dict(
+                type="dict",
+                required=True,
+                options=dict(
+                    device=dict(required=True, type="raw"),
+                    virtual_machine=dict(required=False, type="raw"),
+                    name=dict(required=True, type="str"),
+                    port=dict(required=True, type="int"),
+                    protocol=dict(required=True, type="int"),
```
FragmentedPacket: Protocol can be **raw** so it can accept either the word or integer

plugins/module_utils/netbox_utils.py
TawR1024: Dict already exist, so I only changed the value type.

tests/integration/integration-tests.yml
```diff
@@ -4079,6 +4079,95 @@
...
+          - test_service_update is changed
+
+
+    - name: "NETBOX_SERVICE: Test service deletion"
+      netbox_service:
+        netbox_url: "http://localhost:32768"
+        netbox_token: "0123456789abcdef0123456789abcdef01234567"
+        data:
+          device: "FOR_SERVICE"
+          name: "node-exporter"
```
FragmentedPacket: Thanks for all the changes! Do you mind adding more to the asserts? If you look at each type of test for the other modules, we're attempting to validate some of the data as well. I would just look at a few of the tests for each create, idempotent, update, delete.

Everything else looks OK.

TawR1024: hi, thanks for all your comments.
Test asserts have been updated

tests/integration/integration-tests.yml
```diff
@@ -4018,7 +4018,7 @@
       assert:
         that:
           - test_three is changed
-          - test_three['diff']['after']['upstream_speed'] == 1000
+
```
FragmentedPacket: Can you add this line back? After that I will get this merged.

Thanks for all your work on this! I will try and update the documentation as well to help understand what we're looking for when contributing new modules, etc.

PR #77, opened 2020-01-31
BugFix: Fix device type slug

### LINKED ISSUE #74: netbox_device_type: slug is required
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
The `slug` is used in regions, sites and more places, and is not mandatory there. However, for `netbox_device_type`, it seems to be mandatory. It does not seem that it is necessary since other netbox ansible modules are generating it automatically.

FragmentedPacket:
Fixes #74 

Made **slug** not required, but model is now required. Slug can still be passed in, but if not, it will slugify the model.

### DIFF
--- a/tests/integration/integration-tests.yml
+++ b/tests/integration/integration-tests.yml
@@ -1906,6 +1906,27 @@
           - test_five['device_type'] == None
           - test_five['msg'] == "device_type test-device-type already absent"
 
+    - name: "DEVICE_TYPE 6: Without Slug"
+      netbox_device_type:
+        netbox_url: http://localhost:32768
+        netbox_token: 0123456789abcdef0123456789abcdef01234567
+        data:
+          model: ws-test-3850
+          manufacturer: Test Manufacturer
+        state: present
+      register: test_six
+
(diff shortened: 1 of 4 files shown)


tests/integration/integration-tests.yml
```diff
@@ -1906,6 +1906,27 @@
           - test_five['device_type'] == None
           - test_five['msg'] == "device_type test-device-type already absent"
 
+    - name: "DEVICE_TYPE 6: Without Slug"
+      netbox_device_type:
+        netbox_url: http://localhost:32768
+        netbox_token: 0123456789abcdef0123456789abcdef01234567
+        data:
+          model: ws-test-3850
```
ignatenkobrain: what about to name model `WS Test 3850` to make sure that slugifying works fine?

FragmentedPacket: Yes that is a very good catch. I will update that test this weekend

FragmentedPacket:
@ignatenkobrain Updated. Let me know if this fixes your issue and I'll try and get this merged and a new version out this week with all the fixes to your issues.

PR #97, opened 2020-02-13
Inventory-bug: Changed the lookups to use slug where available

### LINKED ISSUE #59: Bug: Netbox Inventory group_by returns space separated groups
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
When instructing the NetBox inventory to group by device roles, the inventory uses the role's name, which contains spaces.

Original Ansible Core Issues:
https://github.com/ansible/ansible/issues/53380
https://github.com/ansible/ansible/pull/53384

@nikkytub since you've started a PR to use the roles slug, could you expand that to the other `group_by` values and submit a PR to this issue? We've pulled the inventory plugin out of core and into this collection for faster updates.

##### STEPS TO REPRODUCE
The inventory file looks like the following:

```yaml
plugin: netbox
api_endpoint: <endpoint>
validate_certs: no
token: <token>
query_filters:
  - name: QA-DE-1-ASR11a
group_by:
  - sites
  - device_roles
```
After an `ansible-inventory --list`, I expect grouped results by site and role.

##### EXPECTED RESULTS
```json
{
  "all": {
        "children": [
            "device_roles_neutron-router",
            "sites_QA-DE-1a",
            "ungrouped"
        ]
    },
    "device_roles_neutron-router": {
        "hosts": [
            "QA-DE-1-ASR11a"
        ]
    },
    "sites_QA-DE-1a": {
        "hosts": [
            "QA-DE-1-ASR11a"
        ]
    }
}
```
I would expect the inventory to use the role-slug from netbox or the replace the spaces in the role's name by underscores.

##### ACTUAL RESULTS
```json
{
    "all": {
        "children": [
            "device_roles_Neutron Router",
            "sites_QA-DE-1a",
            "ungrouped"
        ]
    },
    "device_roles_Neutron Router": {
        "hosts": [
            "QA-DE-1-ASR11a"
        ]
    },
    "sites_QA-DE-1a": {
        "hosts": [
            "QA-DE-1-ASR11a"
        ]
    }
}
```

FragmentedPacket:
Fixes #59 

This should fix the group_by issue, but may want to be revisited as they Ansible appears to be moving to group_names requiring to use underscores.

This at least fixes the issue above for the time being, but I think Ansible also does the conversions itself.

### DIFF
--- a/plugins/inventory/netbox.py
+++ b/plugins/inventory/netbox.py
@@ -342,23 +342,23 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         url = self.api_endpoint + "/api/dcim/platforms/?limit=0"
         platforms = self.get_resource_list(api_url=url)
         self.platforms_lookup = dict(
-            (platform["id"], platform["name"]) for platform in platforms
+            (platform["id"], platform["slug"]) for platform in platforms
         )
 
     def refresh_sites_lookup(self):
         url = self.api_endpoint + "/api/dcim/sites/?limit=0"
         sites = self.get_resource_list(api_url=url)
-        self.sites_lookup = dict((site["id"], site["name"]) for site in sites)
+        self.sites_lookup = dict((site["id"], site["slug"]) for site in sites)
 
     def refresh_regions_lookup(self):
         url = self.api_endpoint + "/api/dcim/regions/?limit=0"
         regions = self.get_resource_list(api_url=url)
-        self.regions_lookup = dict((region["id"], region["name"]) for region in regions)
+        self.regions_lookup = dict((region["id"], region["slug"]) for region in regions)
 
     def refresh_tenants_lookup(self):
         url = self.api_endpoint + "/api/tenancy/tenants/?limit=0"
         tenants = self.get_resource_list(api_url=url)
-        self.tenants_lookup = dict((tenant["id"], tenant["name"]) for tenant in tenants)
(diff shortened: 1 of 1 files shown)


StepBee:
From my testing it looks like once Ansible will change the default from transform_invalid_group_chars TRANSFORM_INVALID_GROUP_CHARS from the current default "never" to then "always", the netbox inventory plugin will be broken, even with this fix as i think even the dash will be replaced.

The netbox plugin is failing as soon as the group in ansible is named (different because of replacing space or dash with underscore), as the script did try to create the group.

DouglasHeriot:
@StepBee I haven't tested this case, but I think you're right. The code in `add_host_to_groups`:
```
self.inventory.add_group(group=group_name)
self.inventory.add_host(group=group_name, host=hostname)
```
probably will fail if `TRANSFORM_INVALID_GROUP_CHARS` is enabled. The group name passed to `add_host` will not match the group name that was transformed when the group was created.

See `add_group` returns the group name that was actually used.
https://github.com/ansible/ansible/blob/308723c3ca1122363e419070f1fa1d76ff5611a9/lib/ansible/inventory/data.py#L159

I'll address this in my pull request #155 as I'm touching some of this group logic there.

StepBee:
@DouglasHeriot just through your comment i actually noticed, this is for the netbox ansible modules collection and not for the ansible build in netbox inventory plugin - and i have to say i like the netbox ansible module collection a lot, thanks for the work!
I read you are heavily working on #155 these days - the group name cleanup is holding me from already using it - did i understand correct this is not part of the collection yet right?

PR #98, opened 2020-02-13
Inventory-feature: Added interfaces and ip addresses

### LINKED ISSUE #60: Feature: add Device and VM Interface information
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY
Implement a new extractor method which includes the inventory information if requested. Additionally, make a second request to gather all device/vm IP addresses and attach them as a list to the interface object.

Original Ansible Core Issues/PR:
https://github.com/ansible/ansible/pull/55982

@mkeetman, since you submitted the original PR that has yet to be merged into core and we've pulled the inventory out of core and into this collection, could you submit a new PR to this Issue?

FragmentedPacket:
Fixes #60 

This just adds work from PR in Ansible core to address this. All raw data again, but gets it implemented.

### DIFF
--- a/plugins/inventory/netbox.py
+++ b/plugins/inventory/netbox.py
@@ -48,6 +48,12 @@ DOCUMENTATION = """
                 # in order of precedence
                 - name: NETBOX_TOKEN
                 - name: NETBOX_API_KEY
+        interfaces:
+            description:
+                - If True, it adds the device or virtual machine interface information in host vars.
+            default: False
+            type: boolean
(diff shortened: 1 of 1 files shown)


DavidRobbNZ:
Using a dash rather than an underscore makes this a bit of a pain to use in Jinja2 templates; could this change from

                    interface["ip-addresses"] = interface_ip

to:

                    interface["ip_addresses"] = interface_ip

?

PR #116, opened 2020-02-24
WIP: Cable Module for Interface and Circuit connection

TawR1024:


TawR1024:
We need to create new module netboxCables, because NetboxDCIM module use another models based on device names.
Cables model using terminations ids.
So, new module should convert names of device and ports to ids.

FragmentedPacket:
Implemented per #251

PR #120, opened 2020-02-26
Add slug key for sites, tenants and tenant_groups

noonedeadpunk:
This change makes possible for deployers to define slug while site,
tenant or tenant_group creation.

Signed-off-by: Dmitriy Rabotyagov <drabotyagov@vexxhost.com>


FragmentedPacket:
@noonedeadpunk Thanks for your work on this. I've held off on this to try and maintain support for Netbox v2.5, but I'm currently wanting to only support the two latest NetBox releases so we don't have to maintain so much legacy code and I love having the Tenant's be unique with slugs.

I will make a final decision, but leaning towards supporting the two latest releases as I have some workarounds for the choices as well.

PR #126, opened 2020-03-05
BugFix: VC intf fix

### LINKED ISSUE #105: Netbox_interface : Fail to modify interface within a virtual chassis #67578
##### ISSUE TYPE

 - Bug Report

##### SUMMARY
Hi !
I have 2 devices, in a virtual chassis.
virtual_chassis

The interface Ethernet1 is configured on the non-master device

before

On netbox, i see this interface in the virtual-chassis master's interfaces, which is fine.
When i try to modify an interface within a virtual chassis with the netbox_interface module, **it deletes the interfaces from the secondary device and creates it on the master device**
after_slave
after_master

##### STEPS TO REPRODUCE

   - Create two devices in netbox.
   - Gather them with a virtual chassis (choose a master and a slave)
   - Create interface on the slave device
   -  Run the playbook
   -  Notice that the interface disapeared on the slave device and appeared on the master.

```yaml
- name: "Playbook"
  vars :
    nb_addr: "your_nb_address"
    nb_token: "your_token"
  tasks:
    - name :  Create or modify interface
      netbox_interface:
        netbox_url: "{{ nb_addr }}"
        netbox_token: "{{ nb_token }}"
        data:
          device: "ROUTER1"
          name: "Ethernet1"
          description: "Description change"
      delegate_to: 127.0.0.1
```

##### EXPECTED RESULTS

I expect interface on the secondary device to be modified
##### ACTUAL RESULTS

```
ansible-playbook 2.9.4
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  executable location = /usr/bin/ansible-playbook
  python version = 2.7.13 (default, Sep 26 2018, 18:42:22) [GCC 6.3.0 20170516]
Using /etc/ansible/ansible.cfg as config file
setting up inventory plugins
host_list declined parsing /etc/ansible/hosts as it did not pass its verify_file() method
script declined parsing /etc/ansible/hosts as it did not pass its verify_file() method
auto declined parsing /etc/ansible/hosts as it did not pass its verify_file() method
Parsed /etc/ansible/hosts inventory source with ini plugin
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

Loading callback plugin default of type stdout, v2.0 from /usr/lib/python2.7/dist-packages/ansible/plugins/callback/default.pyc

PLAYBOOK: module.yml **********************************************************************************************************************************************************************************************************************************************************
Positional arguments: module.yml
become_method: sudo
inventory: (u'/etc/ansible/hosts',)
forks: 5
tags: (u'all',)
verbosity: 4
connection: smart
timeout: 10
1 plays in module.yml

PLAY [Get port configuration and save in Netbox] ******************************************************************************************************************************************************************************************************************************
META: ran handlers

TASK [Create or modify access interface] **************************************************************************************************************************************************************************************************************************************
task path: /root/netbox_issue/module.yml:12
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: root
<127.0.0.1> EXEC /bin/sh -c 'echo ~root && sleep 0'
<127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /root/.ansible/tmp/ansible-tmp-1582127139.71-92714538077148 `" && echo ansible-tmp-1582127139.71-92714538077148="` echo /root/.ansible/tmp/ansible-tmp-1582127139.71-92714538077148 `" ) && sleep 0'
Using module file /usr/lib/python2.7/dist-packages/ansible/modules/net_tools/netbox/netbox_interface.py
<127.0.0.1> PUT /root/.ansible/tmp/ansible-local-22430U44jhB/tmpsR8lNO TO /root/.ansible/tmp/ansible-tmp-1582127139.71-92714538077148/AnsiballZ_netbox_interface.py
<127.0.0.1> EXEC /bin/sh -c 'chmod u+x /root/.ansible/tmp/ansible-tmp-1582127139.71-92714538077148/ /root/.ansible/tmp/ansible-tmp-1582127139.71-92714538077148/AnsiballZ_netbox_interface.py && sleep 0'
<127.0.0.1> EXEC /bin/sh -c '/usr/bin/python /root/.ansible/tmp/ansible-tmp-1582127139.71-92714538077148/AnsiballZ_netbox_interface.py && sleep 0'
<127.0.0.1> EXEC /bin/sh -c 'rm -f -r /root/.ansible/tmp/ansible-tmp-1582127139.71-92714538077148/ > /dev/null 2>&1 && sleep 0'
changed: [localhost -> 127.0.0.1] => {
    "changed": true,
    "diff": {
        "after": {
            "description": "Description change",
            "device": 1
        },
        "before": {
            "description": "",
            "device": 3
        }
    },
    "interface": {
        "cable": null,
        "connected_endpoint": null,
        "connected_endpoint_type": null,
        "connection_status": null,
        "count_ipaddresses": 0,
        "description": "Description change",
        "device": 1,
        "enabled": true,
        "id": 540,
        "lag": null,
        "mac_address": null,
        "mgmt_only": false,
        "mode": null,
        "mtu": null,
        "name": "Ethernet1",
        "tagged_vlans": [],
        "tags": [],
        "type": "virtual",
        "untagged_vlan": null
    },
    "invocation": {
        "module_args": {
            "data": {
                "description": "Description change",
                "device": 1,
                "name": "Ethernet1"
            },
            "netbox_token": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "netbox_url": "http://192.168.150.100:8000",
            "state": "present",
            "validate_certs": true
        }
    },
    "msg": "Interface Ethernet1 updated"
}
META: ran handlers
META: ran handlers

PLAY RECAP ********************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=1    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

FragmentedPacket:
Fixes #105 

The original attempt at fixing it was nulling out `self.object` instead of `self.nb_object`. I believe this should resolve the issue

### DIFF
--- a/plugins/module_utils/netbox_dcim.py
+++ b/plugins/module_utils/netbox_dcim.py
@@ -93,13 +93,17 @@ class NetboxDcimModule(NetboxModule):
         # This is logic to handle interfaces on a VC
         if self.endpoint == "interfaces":
             if self.nb_object:
-                if self.nb_object.device:
-                    device = self.nb.dcim.devices.get(self.nb_object.device.id)
-                    if (
-                        device["virtual_chassis"]
-                        and self.nb_object.device.id != self.data["device"]
-                    ):
-                        self.nb_object = None
+                device = self.nb.dcim.devices.get(self.nb_object.device.id)
+                if (
+                    device["virtual_chassis"]
+                    and self.nb_object.device.id != self.data["device"]
+                ):
+                    if self.module.params.get("update_vc_child"):
+                        data["device"] = self.nb_object.device.id
+                    else:
+                        self._handle_errors(
+                            msg="Must set update_vc_child to True to allow child device interface modification"
+                        )
 
         if self.state == "present":
             self._ensure_object_exists(nb_endpoint, endpoint_name, name, data)
--- a/plugins/module_utils/netbox_dcim.py
+++ b/plugins/module_utils/netbox_dcim.py
@@ -99,7 +99,7 @@ class NetboxDcimModule(NetboxModule):
                         device["virtual_chassis"]
                         and self.nb_object.device.id != self.data["device"]
                     ):
-                        self.object = None
+                        self.nb_object = None
 
         if self.state == "present":
             self._ensure_object_exists(nb_endpoint, endpoint_name, name, data)
--- a/plugins/modules/netbox_device_interface.py
+++ b/plugins/modules/netbox_device_interface.py
@@ -105,6 +105,13 @@ options:
           - Any tags that the prefix may need to be associated with
         type: list
     required: true
+  update_vc_child:
+    type: boolean
+    default: False
+    description:
+      - |
+        Use when master device is specified for C(device) and the specified interface exists on a child device
+        and needs updated
   state:
     description:
       - Use C(present) or C(absent) for adding or removing.
@@ -190,6 +197,15 @@ EXAMPLES = r"""
           mgmt_only: true
           mode: Tagged
         state: present
+    - name: Update interface on child device on virtual chassis
+      netbox_device_interface:
+        netbox_url: http://netbox.local
+        netbox_token: thisIsMyToken
+        data:
+          device: test100
+          name: GigabitEthernet2/0/1
+          enabled: false
+        update_vc_child: True
 """
 
 RETURN = r"""
@@ -220,6 +236,7 @@ def main():
     argument_spec = NETBOX_ARG_SPEC
     argument_spec.update(
         dict(
+            update_vc_child=dict(type="bool", required=False, default=False),
             data=dict(
                 type="dict",
                 required=True,
@@ -238,7 +255,7 @@ def main():
                     ),
                     untagged_vlan=dict(required=False, type="raw"),
                     tagged_vlans=dict(required=False, type="raw"),
-                    tags=dict(required=False, type=list),
+                    tags=dict(required=False, type="list"),
                 ),
             ),
         )
@@ -247,6 +264,7 @@ def main():
     required_if = [
         ("state", "present", ["device", "name"]),
         ("state", "absent", ["device", "name"]),
+        ("update_vc_child", True, ["device"]),
     ]
 
     module = NetboxAnsibleModule(
--- a/tests/integration/integration-tests.yml
+++ b/tests/integration/integration-tests.yml
@@ -473,7 +473,7 @@
           - test_ten is changed
           - test_ten['msg'] == "interface Ethernet2/1 updated"
           - test_ten['diff']['after']['description'] == 'Updated child interface from parent device'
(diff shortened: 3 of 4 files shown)


Jamboon-beurre:
This does not work as expected : 
Here's a little drawing to explain how it works now : 
issue


This logic adds another problem : after using the module once, two interfaces have the same name on the master device, and then, when you want to use the module again, it says "More than one result returned for Ethernet111", (which is normal)


The logics should be that if an interface is already named 'Ethernet5', it should not create a new one on the master device but modify the existing one on the slave device.

This problem is a bit tricky but thanks for your nice work around this :)

FragmentedPacket:
@Jamboon-beurre What do you think of that?

I can also push up the logic and see how it looks from a diffing point of view as well. Not entirely sure I'm a fan of it, but I can be persuaded.

Does specifying the child create a burden for you? I can possibly see how it can if the stack is large enough or maybe the user is not sure which switch in stack the interface belongs to

Jamboon-beurre:
Well, i'm not sure what to think about this :
It would be easier for me not to specify the child : since the interface is shown on the parent, it seems logic to be able to modify it with the parent's name.
But on the other hand, it may lead to confusion to update an interface of a device using the name of another device.
Maybe we can try to push this to test it  and maybe to get feedbacks from other users on this issue

FragmentedPacket:
@Jamboon-beurre I will try and get this PR updated to account for this use case (updating via the parent) and I will post back once it's implemented so you can test.

FragmentedPacket:
@Jamboon-beurre Ok so I pushed up an update for an option that can be toggled: `update_vc_child` that will update a child interface if the parent/master device is specified for `device` and the interface exists.

The flow of this change:
1. Parent/master device is specified, it finds the interface, but if the device ID is not the same as the parent/master device passed in (interface belongs to child device) and the parent device is a `virtual_chassis` it will then check whether `update_vc_child` is true or false. If true, it will update the device to the child ID so it won't move the interface over to the master. If false, it will fail and advise user to set the var to True.

2. If the interface doesn't exist, it will create it on the device specified - the parent/master (normal behavior)

Some confusion that may happen is we let someone update the interface from parent view by specifying the parent device as long as the interface exists, but if the interface does not exist, it will create it under the parent device. With this, we assume the user knows the interface exists already.

Jamboon-beurre:
I've tested it and it works fine.
I think it's a good idea to add a variable to choose how to deal within this situation.
Thank you for your work on this problem !

Issue #45, opened 2019-12-03
Can not assign ip-address to virtual-machine


inetman28:
##### ISSUE TYPE
 - Module question

##### SUMMARY
Can not assign ip-address to virtual-machine interface. 

##### STEPS TO REPRODUCE
You can create following play to test behavior:
```
    - name: assign ip to vm
      netbox_ip_address:
        netbox_url: "{{ nb_url }}"
        netbox_token: "{{ nb_token }}"
        data:
          interface:
            name: ens38
            virtual_machine: "{{ vm_name }}"
          address: 10.20.30.40/24
        validate_certs: no
```
##### EXPECTED RESULTS

I expect that ip-address 10.20.30.40 will assign to virtual-machine interface 'ens38' with ip-address 10.20.30.40/24. How I can to do this? 

##### ACTUAL RESULTS
The task failed:
```
TASK [assign ip to vm] ****************************************************************************************************************************
[...]

PR #132, opened 2020-03-06
BugFix: Jinja template always passes in string, allowed find_ids method to account for that

FragmentedPacket:
Fixes a reported issue in #45 

When using Jinja templating, Ansible converts to string. Added an attempt to convert the value passed in to an integer


Duck-dave:
It works! Great Job, 
#45

PR #136, opened 2020-03-15
Added flag for raw api output

malbertus:
This PR adds the ability to return the raw data of the Netbox API. 

Currently when given the command `device: "{{ lookup('netbox.netbox.nb_lookup', 'devices', api_endpoint=api_url, token=api_token }}"` data is returned in the following format:

```
    "device": {
            "key": 7,
            "value": {
                "asset_tag": null,
                "cluster": null,
                "comments": "",
                <snipped for brevity>
      }
```
This adds an extra layer of complexity for usage as it increases the depth of the data structure behind a somewhat meaningless key/value construct where the Netbox object id becomes the key and the payload is hidden under value

This PR adds the option for the lookup plugin to return the raw Netbox API response in the following format

`device: "{{ lookup('netbox.netbox.nb_lookup', 'devices', api_endpoint=api_url, token=api_token, raw_data=True }}"`

```
    "device": {
        "asset_tag": null,
        "cluster": null,
        "comments": "",
        <snipped for brevity>
      }
```

Items with multiple returns will be available in a list format

```
    "interfaces": [
        {
            "cable": {
                "id": 1,
                "label": "",
                <snipped for brevity>
        },
        {
            "cable": {
                "id": 2,
                "label": "",
                <snipped for brevity>
         }
   ]
```

### DIFF
--- a/plugins/lookup/nb_lookup.py
+++ b/plugins/lookup/nb_lookup.py
@@ -233,7 +233,7 @@ class LookupModule(LookupBase):
                     if netbox_raw_return:
                         results.append(dict(res))
 
-                    else: 
+                    else:
                         key = dict(res)["id"]
                         result = {key: dict(res)}
                         results.extend(self._flatten_hash_to_list(result))
--- a/plugins/lookup/nb_lookup.py
+++ b/plugins/lookup/nb_lookup.py
@@ -52,6 +52,10 @@ DOCUMENTATION = """
             description:
                 - The location of the private key tied to user account.
             required: False
+        raw_data:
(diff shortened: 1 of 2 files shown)


FragmentedPacket:
Thanks for taking care of this! You can get the tests to pass by `pip install black` and then running `black .` or `black plugins/lookup/nb_lookup.py` for it to format the file.

Can you also add integration tests for this new option? You can just add one that has this new option as `True` and then assert the data should be the way it is. If you have any questions on how to structure the tests then hit me up on Slack or here.

PR #145, opened 2020-03-27
Update Travis test environments to Netbox 2.6, 2.7

DouglasHeriot:
Still a work in progress. Running into issues with Ansible 2.10 dev - ipaddress compat class for Python 2.7 hosts has been moved to the netcommon collection. Need to work out how this import should work now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,8 +57,13 @@ before_script:
   - ansible-galaxy collection install $COLLECTION_NAMESPACE-$COLLECTION_NAME-$COLLECTION_VERSION.tar.gz -p /home/travis/.ansible/collections
 
 script:
-  - ansible-test units --python $PYTHON_VER -v
-  - black . --check
+  # Perform tests on collection from within the installed directory, not the source directory
+  # Required for imports of other collections (ie. ansible.netcommon) to work correctly
+  - (cd /home/travis/.ansible/collections/ansible_collections/$COLLECTION_NAMESPACE/$COLLECTION_NAME && ansible-test units --python $PYTHON_VER -v)
+
+  # Check python syntax
+  - black . --check --diff
+
   - timeout 300 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' localhost:32768)" != "200" ]]; do echo "waiting for Netbox"; sleep 5; done' || false
   - python tests/integration/netbox-deploy.py
   - ansible-playbook tests/integration/regression-tests.yml -vvvv
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,13 +14,13 @@ env:
 
 matrix:
   include:
-    - name: "Python 3.6 - Netbox 2.5 - Latest PyPi Ansible"
+    - name: "Python 3.6 - Netbox 2.6 - Latest PyPi Ansible"
       python: 3.6
-      env: PYTHON_VER=3.6 VERSION=v2.5
+      env: PYTHON_VER=3.6 VERSION=v2.6
       install:
         - cd ..
         # Setup netbox container for integration testing
-        - git clone https://github.com/FragmentedPacket/netbox-docker.git
+        - git clone https://github.com/netbox-community/netbox-docker.git
         - cd netbox-docker
         - docker-compose pull
         - docker-compose up -d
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -12,15 +13,15 @@ env:
...
-    - name: "Python 3.6 - Netbox 2.5 - Latest PyPi Ansible"
+    - name: "Python 3.6 - Netbox 2.6 - Latest PyPi Ansible"
       python: 3.6
-      env: PYTHON_VER=3.6 VERSION=v2.5
+      env: PYTHON_VER=3.6 VERSION=v2.6
       install:
         - cd ..
         # Setup netbox container for integration testing
-        - git clone https://github.com/FragmentedPacket/netbox-docker.git
+        - git clone https://github.com/netbox-community/netbox-docker.git
```
FragmentedPacket: If you change this back, it should work for NetBox 2.6. There is something funky with the original repo which is why I cloned it. I believe it had to do with something crashing during startup scripts.

You can keep the other one the same for 2.7 since that appears to work properly.

.travis.yml
```diff
@@ -56,8 +57,17 @@ before_script:
...
-  - ansible-test units --python $PYTHON_VER -v
-  - black . --check
+  # Perform tests on collection from within the installed directory, not the source directory
+  # Required for imports of other collections (ie. ansible.netcommon) to work correctly
+  - (cd /home/travis/.ansible/collections/ansible_collections/$COLLECTION_NAMESPACE/$COLLECTION_NAME && ansible-test units --python $PYTHON_VER -v)
+
+  # Check python syntax
+  - black . --check --diff
+
+  # Just testing what's up with these containers at this point
```
FragmentedPacket: Won't need these after addressing first review comment

.travis.yml
```diff
@@ -3,7 +3,8 @@ language: python
 services:
   - docker
 
-sudo: false
+os: linux
```
FragmentedPacket: Is there a benefit to changing the container?

.travis.yml
```diff
@@ -66,7 +76,6 @@ script:
 
 deploy:
   provider: script
-  skip_cleanup: true
```
FragmentedPacket: I believe this should remain as we need to build the `tar.gz` to ship to Ansible Galaxy and will fail on release if this option is removed.

.travis.yml
DouglasHeriot: Yeah ok, that has fixed it.
FYI the reason for using the official one in 2.7 is to meet the new Redis requirements.

.travis.yml
DouglasHeriot: A warning in Travis appeared saying "root: missing os, using the default linux". Sounds like a good idea to specify this explicitly. 
I also specified the distribution to use Bionic (18.04) instead of the default of Xenial (16.04). I don't see any reason not to upgrade, and allows for testing the latest version of Python 3.8.

.travis.yml
DouglasHeriot: Travis build config validation is telling me "deploy: deprecated key skip_cleanup (not supported in dpl v2, use cleanup)". 

However I'm confused by this message now - it looks like it only applies to the travis dplv2 which isn't the default yet? Maybe I'll leave this in in-case it's still actually required in dplv1?
https://docs.travis-ci.com/user/deployment-v2#cleaning-up-the-git-working-directory

DouglasHeriot: I've added this line back in for now, as I haven't tested if the Travis deprecation warning is correct or not.

DouglasHeriot:
Ok I think this is ready to merge now.

One significant change was in f95dd89d5eaf652d3ef391e27c7ff033c80ecf11 where I added a third build environment.

The previous test environments:
1. Previous netbox, latest ansible
2. Latest netbox, dev ansible

This left the most-used combination of latest netbox with latest ansible untested.

New test environments:
1. Previous netbox, latest ansible
2. Latest netbox, latest ansible
3. Dev netbox, dev ansible, Python 3.8

This way if the dev ansible is broken for any reason we're still getting a test against the latest netbox. Also introduces testing against the dev version of netbox so we can catch any potential issues there.

Build 2 (Netbox 2.7, latest Ansible) is currently failing - I believe just due to #149 

For now I've told Travis to allow build 3 to fail. There's a few issues with dev ansible that might require a bit more time to investigate and fix. I'm thinking we should try and keep this pull request small and fix those issues separately?

FragmentedPacket:
@DouglasHeriot Thanks for this! I will try and review tomorrow. Work and this whole COVID-19 stuff has made my schedule pretty hectic, but just wanted to let you know I appreciate all your work on this project!

PR #147, opened 2020-03-29
Enhancement: Added option to not prepend group names with group_by option

### LINKED ISSUE #138: Remove or provide option to NOT append the "group_by" group name to inventory groups
##### ISSUE TYPE

 - Feature Idea

##### SOFTWARE VERSIONS
all
###### Ansible:

all
###### Netbox:

all
###### Collection:

all
##### SUMMARY

When using the inventory plugin and specifically the `group_by` option the group_by name is prepended to the groups that are created for Ansible.  If I have device that has a role of branch_switch and I am grouping by `device_role`, in ansible the group is now `device_role_branch_switch` and now all of my group_vars have to be renamed to support this.
##### STEPS TO REPRODUCE

Instead of the inventory group names having the group_by prepended it would omit those.

```yaml
---
api_endpoint: https://netbox
token: TOKEN
validate_certs: True
config_context: True
group_by:
  - sites
  - device_roles
```

##### EXPECTED RESULTS

I would expect that if I group by `device_roles`, or any other option, I would just get the name (i.e. `branch_switch`)
##### ACTUAL RESULTS

What I get is the groups have the group_by variable prepended to groups `device_role_branch_switch`

```

```

FragmentedPacket:
Fixes #138 

Added `group_names_raw` to provide the group name slug rather than prepending with the group_by option.

e.g. Before
```
group_by:
  - device_roles

device_roles_core-switch
```
e.g. After
```
group_by:
  - device_roles

core-switch
```

This also replaces any dash in the slug to an underscore since Ansible will be enforcing that moving from 2.10+. I'm open to keeping it as just returning the slug so please discuss what you guys think is the best approach.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -606,7 +606,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 continue
 
             for sub_group in sub_groups:
-                sub_group = sub_group.replace('-', '_')
+                sub_group = sub_group.replace("-", "_")
                 if self.group_names_raw:
                     group_name = sub_group
                 else:
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -606,6 +606,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 continue
 
             for sub_group in sub_groups:
+                sub_group = sub_group.replace('-', '_')
                 if self.group_names_raw:
                     group_name = sub_group
                 else:
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -67,6 +67,11 @@ DOCUMENTATION = """
                 - manufacturers
                 - platforms
             default: []
+        group_names_raw:
+            description: Will not add the group_by choice name to the group names
+            default: False
+            type: boolean
+            version_added: "0.2.0"
         query_filters:
             description: List of parameters passed to the query string (Multiple values may be separated by commas)
             type: list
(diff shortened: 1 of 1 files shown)


FragmentedPacket:
@smolz @DouglasHeriot @TawR1024 @Yannis100

Can I get feedback from y'all on this change?

Yannis100:
Thanks for asking!
I would rather prefer an option to include prefix or not, and by default have it like before
The option to have the prefix shrinked/shortened to a specific length could be great too, it's the way I implemented it in my "enhanced" version of the inventory plugin

DouglasHeriot:
My concern with this is that without the prefixes, all the slugs share the same namespace. However I can't think of any concrete examples where this will create an issue (ie. a manufacturer with the same name as a site)

As long as there's an option to turn it on/off I think this is ok.

I think in our usage of this we might leave the prefixes on as the verbosity helps when reading configurations to understand what's happening. And when running `ansible-inventory` to see what's there it is a lot clearer when sorted by all the site groups together, all the manufacturer groups together, etc.

The group names "manufacturers_apple" and "platforms_apple_ios" are a lot clearer than "apple" and "apple_ios" - I can much more easily explain to colleages where to find this thing in netbox.

Another thought - there's no reason that both sets of groups couldn't co-exist. Could help people transition, but would also increase confusion.

ThomasADavis:
We use the group names to build up a genders file - and with the roles prefixed, that file gets ugly quite fast.

as for the '-' to '_', ansible is being stupid in my mind..  I'd like a flag to keep them not converted (we already do the ansible.cfg flag to turn  off the warning)

FragmentedPacket:
Alright. It seems like keeping the way it does it now should stay long term (forever) and adding the new option would also stick around and not replace the current method. I will also remove any data manipulation when it comes to changing from a dash to underscore as Ansible provides the config option to ignore the warnings.


@Yannis100 Do you mind providing me an example of the shortening?

Yannis100:
You can find it in this branch of my fork
https://github.com/Yannis100/ansible/tree/region-group-netbox
https://github.com/ansible/ansible/issues/60124
Haven't work on ansible for a while so not sure if it's updated with upstream

DouglasHeriot:
@Yannis100 I have to say thanks for your fork - I've been using your inventory for a little while and am now working on moving some features (the ones I need at least) into this project. I'll be doing regions next...

Issue #150, opened 2020-03-30
netbox_services missing tags, ip address support


ThomasADavis:
##### ISSUE TYPE
 - Bug Report
 - Feature Idea

###### Ansible:
ansible 2.9.6
  config file = /home/tdavis/netbox/ansible.cfg
  configured module search path = ['/home/tdavis/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/tdavis/.local/lib/python3.6/site-packages/ansible
  executable location = /home/tdavis/.local/bin/ansible
  python version = 3.6.8 (default, Nov 21 2019, 19:31:34) [GCC 8.3.1 20190507 (Red Hat 8.3.1-4)]

###### Netbox:
netbox v2.7.11

###### Collection:
v 1.1

##### SUMMARY

netbox 2.7 services can have tags, and IP addresses assigned to them.

the netbox services ansible plugin does not support these two features.

ThomasADavis:
Tags is a 1 line fix..

As for the IP addresses..  I figured it out, you have to give it the ipaddresses ID for it to work, which sucks.  I'd rather have it look them up..  hmm..

FragmentedPacket:
For the IP addresses, that shouldn't be too hard to do. I believe we just need to add `ipaddresses` to a few spots in `netbox_utils` (mimic ip_addressesfor it to be able to properly search for the ip addresses provided within the list.

This is how most of the data resolution currently works to provide the ID to requests to NetBox.

ThomasADavis:
Ok, got this to work:

```
---
- hosts:
    - put-a-host-name-here 
  collections:
    - netbox.netbox
  vars:
    services:
      - name: pdns-auth
        port: 53
        protocol: TCP
        description: PowerDNS authoritative service
        ipaddresses: 
          - address: 192.168.83.4

      - name: pdns-auth
        port: 53
        protocol: UDP
        description: PowerDNS authoritative service
        ipaddresses: 
          - address: 192.168.83.4
        tags:
          - primary
  tasks:
  - name: create services
    netbox_service:
      netbox_url: https://localhost
      netbox_token: redacted
      data:
        device: "{{ ansible_hostname }}"
        name: "{{ item.name }}"
        port: "{{ item.port }}"
        protocol: "{{ item.protocol }}"
        description: "{{ item.description | default(omit) }}"
        tags: "{{ item.tags | default(omit) }}"
[...]

PR #152, opened 2020-04-01
Netbox services fixes issue #150

ThomasADavis:
For issue #150 

This adds/fixes:

support for tags
lookup IP addresses, instead of passing the IP address ID
remove extra 'services' from an dict
add 'protocol' to the service query, which allows proper updates.


FragmentedPacket:
I think adding the protocol to the `ALLOWED_QUERY_PARAMS`  messes with the way it looks up and is causing the integration tests to fail.

I just took a quick glance, but hopefully I'll have time to look within the next week

Issue #143, opened 2020-03-25
Inventory Optimisation: Querying for services and interfaces in bulk rather than by device


DouglasHeriot:
##### ISSUE TYPE

 - Feature Idea

##### SUMMARY

Right now when the inventory plugin fetches the services and interfaces associated with devices and VMs, it does a separate request for each device's services and interfaces. In some circumstances it could be more efficient to just query for all interfaces and all services (similar to how everything else is queried) and then match them up with each device the whole list has been downloaded.

This method would be more efficient in the case where the entire contents of Netbox is wanted in the Ansible inventory. The down side of this would be if there are query_filters in place the limit what devices are required - there would be no way to filter the services or interfaces by device.

Right now I've got 2000 devices in Netbox, less than 1000 services defined, and 10,000 interfaces. And this is less than 10% of our current infrastructure - we're only getting started putting stuff into Netbox. This currently requires 4000 HTTP requests to get the services and interfaces of each device, when it could be achieved with just a couple of batch requests. I haven't even looked at what sort of server CPU and database load this creates.

Alternatively - should I make a request with the Netbox project to include returning interfaces and services as part of the main /dcim/devices/ query?

##### EXPECTED RESULTS

```
Fetching: https://netbox/api/ipam/services/?limit=0
```

##### ACTUAL RESULTS

```
Fetching: https://netbox/api/ipam/services/?device=HIL-01
Fetching: https://netbox/api/ipam/services/?device=HIL-02
Fetching: https://netbox/api/ipam/services/?device=HIL-03
Fetching: https://netbox/api/ipam/services/?device=HIL-04
Fetching: https://netbox/api/ipam/services/?device=HIL-05
Fetching: https://netbox/api/ipam/services/?device=HIL-06
Fetching: https://netbox/api/ipam/services/?device=HIL-07
Fetching: https://netbox/api/ipam/services/?device=HIL-08
...
```

FragmentedPacket:
Could you provide any numbers to this? Possibly print time before and after the fetches or maybe the complete run of this between the two options?

Thinking about this quickly, maybe add if/else to either fetch all or per device depending on the `query_filters`? If no query filters, and we want to fetch those, then it would make sense to just fetch everything with the `limit=0`

I don't have a big data set to test the change against on whether or not it's worth it to implement it.

I'll look for more discussion on this and the PR when it's put in. Don't have a strong opinion on it, but would like to see if there is a performance improvement.

[...]

PR #154, opened 2020-04-03
Inventory: make it optional to query for services (#143)

DouglasHeriot:
Querying every single device's services as a separate HTTP request can be very slow. Allow users to disable this (similar to interfaces) if it is not required.

This makes the inventory usable again, until #143 can be implemented fully

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -54,6 +54,12 @@ DOCUMENTATION = """
             default: False
             type: boolean
             version_added: "0.1.7"
+        services:
+            description:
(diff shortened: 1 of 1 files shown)


plugins/inventory/nb_inventory.py
```diff
@@ -54,6 +54,12 @@
             default: False
             type: boolean
             version_added: "0.1.7"
+        services:
+            description:
+                - If True, it adds the device or virtual machine services information in host vars.
+            default: True
+            type: boolean
+            version_added: "0.1.11"
```
FragmentedPacket, line 62: Next version is 2.0 :)

Were you going to make this a single query?

PR #153, opened 2020-04-02
Inventory: Separate query_filters for devices and vms #140

DouglasHeriot:
Implement separate `device_query_filters` and `vm_query_filters` as described in #140 to allow for filtering on parameters specific to devices or vms, and still continue to fetch the other type.

Also started on #56 some unit tests for the inventory, specifically to test these changes.

Issue #141, opened 2020-03-25
Change request: Remove unnecessary arrays in inventory host vars


DouglasHeriot:
##### ISSUE TYPE

 - Feature Idea

##### SUMMARY

I want to echo the comment by @Yannis100 
https://github.com/ansible/ansible/pull/60642#issuecomment-528259999

> I was wondering why values in vars like config_context, sites, manufacturers and so on, were contained in [] brackets; they can have only 1 value and we have to add [0] every time we want to access them. I know that changing this means a lot of people will have to adapt their templates/playbooks but I don't get why it was done like this in the first place.

I guess it just has to do with how `group_extractors` have 2 separate purposes - not just populating groups, but also filling in host variables. The caller `add_host_to_groups` assumes extractors always output an array - but wouldn't be hard to change that.

It would also be nice to rename the host variables to be singular instead of plural as necessary - eg. "platform" instead of "platforms". There is no way in the Netbox database for there to be more than one platform of a device.

I plan to implement a fix for this - are we happy for this to be a breaking change, or should it have an option to switch this on/off?

##### EXPECTED RESULTS

```
$ ansible-inventory --graph --vars
...
  |--SOME_DEVICE
...
  |  |--{config_context = {}}
  |  |--{device_role = 'ethernet-switch'}
  |  |--{device_type = 'ers-4850gts-pwr'}
  |  |--{manufacturer = 'extreme-networks'}
  |  |--{platform = 'boss'}
  |  |--{rack = 'SOME_RACK'}
  |  |--{site = 'SOME_SITE'}
  |  |--{tags = ['Ansible']}
```

##### ACTUAL RESULTS

```
$ ansible-inventory --graph --vars
...
  |--SOME_DEVICE
...
  |  |--{config_context = [{}]}
  |  |--{device_roles = ['ethernet-switch']}
  |  |--{device_types = ['ers-4850gts-pwr']}
  |  |--{manufacturers = ['extreme-networks']}
  |  |--{platforms = ['boss']}
  |  |--{racks = ['SOME_RACK']}
  |  |--{sites = ['SOME_SITE']}
  |  |--{tags = ['Ansible']}
```

FragmentedPacket:
I agree with these changes as well.

I'd say this may have a greater impact so probably have a toggle for it, and then we can deprecate that option eventually.

FragmentedPacket:
Closed via #155

Issue #56, opened 2020-01-11
Add Tests: Netbox Inventory Plugin


FragmentedPacket:
##### ISSUE TYPE

 - Testing Update

##### SUMMARY
Need to build unit tests and possibly some integration testing for Netbox inventory plugin

DouglasHeriot:
I plan to have a look at this as part of my other work on the inventory plugin.

Going to look at structuring the integration tests in such a way we can use the `ansible-test` tool as we do for the unit tests.
https://docs.ansible.com/ansible/latest/dev_guide/testing_integration.html

FragmentedPacket:
@DouglasHeriot Let me know how that goes. Last time I checked it doesn't allow Docker compose which is the way we're building NetBox currently for integration testing. I wouldn't spend too much time on it if it becomes too much of a burden since we're already spinning up an instance for integration tests.

DouglasHeriot:
We can probably close this now, implemented by #155 ?
It's not at 100% code coverage, but it's pretty solid.

FragmentedPacket:
Agreed.

PR #155, opened 2020-04-03
Inventory: Remove unnecessary lists around singular host vars #141

DouglasHeriot:
Implements #141 
There's a new `plurals` option on the inventory. It is `True` by default to maintain backwards compatibility. When set to `False`, the `group_by` options and `host_vars` all change names to become singular, and unnecessary lists around singular vars are removed.

Once #153 is merged to start on #56 inventory unit tests, I'll add some unit tests for `plurals` too.


DouglasHeriot:
This is still a work in progress, not actually ready to merge. Need to do a lot more testing.

DouglasHeriot:
This is getting closer. Currently working on #56 adding integration tests as part of this to ensure the work on the `plurals` option doesn't cause any regressions.

I've found the `community.aws` collection has a good example of how to do this. https://github.com/ansible-collections/community.aws/tree/master/tests/integration/targets/script_inventory_ec2

Got caught for a little bit on ansible/ansible#68963

DouglasHeriot:
Ok, @FragmentedPacket this is finally ready for a proper review.
Sorry it's become so large - the `plurals` option is a substantial change so I wanted to submit tests. Ended up going down the [rabbit hole](https://www.ansible.com/hubfs//AnsibleFest%20ATL%20Slide%20Decks/Quality%20Matters_%20Improving%20your%20Ansible%20modules%20and%20plugins%20with%20ansible-test.pdf) of improving all of the integration tests along the way.

Here's a summary of what this pull request includes:
* #141 Introduce `plurals` option to allow removing unnecessary arrays, and rename vars/groups with correct singular/plural form
* #56 Implement integration testing for the inventory plugin. The test is checking the output of `ansible-inventory --list` against known correct output
* #97 Fix compatibility with Ansible's TRANSFORM_INVALID_GROUP_CHARS = always setting.
* #178 Implement `region` `group_by` option, and extra `regions` as `host_vars`
* A small step towards #143 performance improvement - don't call extract_interfaces, ipaddresses and services twice for every host

More details:
* `plurals` defaults to `True` for backwards compatibility. (Tested by `test-inventory-legacy.yml`). Eventually we might want to add a warning and make the default `False` as it makes more sense.
* When `plurals` is `False`, the host var `device_role` has been renamed to `role` to match `role` on VMs.
* `CONTRIBUTING.md` has been updated with how I've set up my test environment. It also has details of how to update the inventory test output.
* Existing integration tests for the modules have been moved into the `tests/integration/targets` directory, and changed slightly for compatibility with the `ansible-test` tool. The main benefit I see of this is we get tooling like code coverage checking sorted out automatically. As these are run like Ansible roles and not playbooks, the `netbox.netbox` prefix has to be added to all module calls.
* Added travis testing against Netbox 2.8 with release version of Ansible. Stuck to Python 3.7 instead of 3.8 due to ansible/ansible#67118 Using Netbox `develop` branch `snapshot` docker image due to netbox-community/netbox#4496

.editorconfig
```diff
@@ -0,0 +1,11 @@
+
```
FragmentedPacket, line 1: Just starting to review, but do you mind getting rid of this file?

FragmentedPacket:
@DouglasHeriot I will try and get to this during the weekend. Work has been kicking my butt so haven't had the mental capacity to walk through the PR.

Thanks for all your effort!

.editorconfig
DouglasHeriot: Yeah I can remove it if you'd like - just thought it would be helpful and was a fairly standard thing.

plugins/inventory/nb_inventory.py
```diff
@@ -50,6 +50,12 @@
                 # in order of precedence
                 - name: NETBOX_TOKEN
                 - name: NETBOX_API_KEY
+        plurals:
+            description:
+                - If True, all host vars are contained inside single-element arrays for legacy compatibility. Group names will be plural (ie. "sites_mysite" instead of "site_mysite")
+            default: True
+            type: boolean
+            version_added: "0.1.11"
```
FragmentedPacket, line 58: Update version to `0.2.1`

plugins/inventory/nb_inventory.py
```diff
@@ -461,23 +476,52 @@ def get_resource_list(self, api_url):
...
-            "platforms": self.extract_platform,
-            "device_types": self.extract_device_type,
-            "services": self.extract_services,
-            "config_context": self.extract_config_context,
-            "manufacturers": self.extract_manufacturer,
-            "interfaces": self.extract_interfaces,
-            "custom_fields": self.extract_custom_fields,
-        }
+
+        if self.plurals:
```
FragmentedPacket, line 480: Do we need this logic here? It appears each of these functions call the `pluralize` function to then return a list or just the value.

plugins/inventory/nb_inventory.py
```diff
@@ -638,6 +684,39 @@ def extract_custom_fields(self, host):
...
+            # Device has no site
+            return []
+
+        regions = []
+        region_id = self.sites_region_lookup[site_id]
+
+        # Keep looping until the region has no parent
+        while region_id is not None:
+            region_slug = self.regions_lookup[region_id]
+            if region_slug in regions:
```
FragmentedPacket, line 710: I tested this and appears it doesn't get stuck in a loop, but maybe my test data isn't correct for L710-L712

.travis.yml
```diff
@@ -80,31 +101,48 @@ before_script:
...
   - black . --check --diff
 
+  # Sanity tests
+  # --requirements - install pip packages as necessary
+  # Skip pep8 as we use black instead
+  # Skip validate-modules until the issues can be resolved #180
+  - ansible-test sanity -v --requirements --python $PYTHON_VER --skip-test pep8 --skip-test validate-modules
+
+  # Unit tests, with code coverage
+  - ansible-test units -vvvv --coverage --python $PYTHON_VER
```
FragmentedPacket, line 125: After looking at the tests, what is the least amount of verbosity can we get away with these commands? The test output is pretty heavy with setup information for all this.

plugins/inventory/nb_inventory.py
```diff
@@ -461,23 +476,52 @@ def get_resource_list(self, api_url):
...
+                "device_type": self.extract_device_type,
+                "services": self.extract_services,
+                "config_context": self.extract_config_context,
+                "manufacturer": self.extract_manufacturer,
+                "interfaces": self.extract_interfaces,
+                "custom_fields": self.extract_custom_fields,
+                "region": self.extract_regions,
+            }
+
+    def _pluralize(self, something):
```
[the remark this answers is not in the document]
DouglasHeriot: That's fair, I'll change this to extracted_value

plugins/inventory/nb_inventory.py
DouglasHeriot: The keys in the dictionary are different for each option - determines what the `host_vars` will be named, and what `group_by` options are valid.

plugins/inventory/nb_inventory.py
DouglasHeriot: No this will never get stuck in a loop – wasn't sure how defensively I should be programming against the Netbox API. It would only be in theory if Netbox had a corrupt database that somehow had a loop in it.

.travis.yml
DouglasHeriot: Yeah it's very verbose. I wasn't sure whether to go for enough information to debug any issues, or just enough to read and see where it failed. These could probably all be changed to just `-v` or `-vv`

plugins/inventory/nb_inventory.py
FragmentedPacket: Ok that makes sense.

plugins/inventory/nb_inventory.py
FragmentedPacket: Ok no problem. I don't feel strongly on either having it or not so we can leave it.

.travis.yml
FragmentedPacket: Yes that is definitely something to consider. I'll leave that decision up to you.

DouglasHeriot: I've changed it to `-v`. Should be enough for most cases.

plugins/inventory/nb_inventory.py
DouglasHeriot: I've left this in but updated the comment to be more clear.

plugins/inventory/nb_inventory.py
DouglasHeriot: Done

FragmentedPacket:
@DouglasHeriot Thanks a lot for this. I can't tell you how much I appreciate your work on this. Great work and the quality is superb.

DouglasHeriot:
@FragmentedPacket thanks for being prompt and easy to work with!
(The Ansible 2.10 release is going to come with a bit of pain, but being able to actually get stuff like this merged with solid testing is big win)

PR #168, opened 2020-04-08
Update rack width to comply with latest API documentation

### LINKED ISSUE #167: DCIM/Rack: width is changed on every run
##### ISSUE TYPE
 - Bug Report

###### Ansible:
2.9.6

###### Netbox:
2.7.11

###### Collection:
devel

##### SUMMARY
The value for the rack width was wrongly documented as str, while it should be int.
See https://github.com/netbox-community/netbox/issues/4274 for more details.
Because of the wrong documentation the rack width is expected as str in the `netbox_rack` module.

This leads to constant changes on rack objects with explicitly defined rack width on every run.

##### STEPS TO REPRODUCE
Run the following playbook twice:
```yaml
---
- hosts: localhost
  connection: local
  gather_facts: False
  collections:
    - netbox.netbox
  tasks:
  - name: Create Rack(s)
    netbox_rack:
      netbox_url:
      netbox_token:
      data:
        name: Test Rack
        site: Test Site
        width: 19
        status: Planned
      state: present
```

##### EXPECTED RESULTS
First run creates the rack.
Second run says OK and does not change anything.

##### ACTUAL RESULTS
First run creates the rack.
Second run changes the rack width from 19 to "19".
```
--- before
+++ after
@@ -1,3 +1,3 @@
 {
-    "width": 19
+    "width": "19"
 }
```

toerb:
This fixes #167 and adds a new rack width choice (10 inches) introduced in NetBox 2.7.9 (https://github.com/netbox-community/netbox/issues/4305).

### DIFF
--- a/tests/integration/integration-tests.yml
+++ b/tests/integration/integration-tests.yml
@@ -1549,7 +1549,7 @@
           - test_five['diff']['after']['tags'][0] == "Schnozzberry"
           - test_five['diff']['after']['type'] == 100
           - test_five['diff']['after']['u_height'] == 48
-          - test_five['diff']['after']['width'] == "23"
+          - test_five['diff']['after']['width'] == 23
           - test_five['rack']['name'] == "Test rack one"
           - test_five['rack']['site'] == 1
           - test_five['rack']['asset_tag'] == "1234"
@@ -1566,7 +1566,7 @@
           - test_five['rack']['tags'][0] == "Schnozzberry"
           - test_five['rack']['type'] == 100
           - test_five['rack']['u_height'] == 48
-          - test_five['rack']['width'] == "23"
+          - test_five['rack']['width'] == 23
           - test_five['msg'] == "rack Test rack one updated"
 
     - name: "6 - Create rack with same asset tag and serial number"
--- a/tests/integration/integration-tests.yml
+++ b/tests/integration/integration-tests.yml
@@ -1598,7 +1598,24 @@
     - name: "6 - ASSERT"
       assert:
         that:
-          - test_six is succeeded
+          - test_six is not changed
+          - test_six['rack']['name'] == "Test rack one"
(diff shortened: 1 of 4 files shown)


tests/integration/integration-tests.yml
```diff
@@ -1566,7 +1566,7 @@
           - test_five['rack']['tags'][0] == "Schnozzberry"
           - test_five['rack']['type'] == 100
           - test_five['rack']['u_height'] == 48
-          - test_five['rack']['width'] == "23"
```
FragmentedPacket, line 1569: Do you mind adding or updating one of the tests to account for this bug? So have the rack created/updated with the width and then update it again to make sure it is still idempotent?

toerb:
Is the last commit good enough for the testing part, @FragmentedPacket?

FragmentedPacket:
I would add `not changed` just to make sure it didn't change and then maybe add other asserts of what it should look like. You can take mostly the same asserts from the previous task to validate the data is the same.

```
- test_eight not changed
- test_eight['rack']['width'] == 23
- etc. from the previous task
```

toerb:
Isn't succeeded implying that it did not change? I was not aware that I could test for not changed (that's what I initially wanted), since I did not find any documentation on that after a quick search.

Either way, after thinking about it I came up with checking for success, since that is what I would 
like to see as a user of this module.

FragmentedPacket:
As long as ignore_errors isn't set, it should fail if it didn't succeed, but I assume it would mark it as succeeded even if it didn't change.

We're asserting that it didn't change rather than succeeded (that is assumed) or else the playbook crashes. At least that is what I understand of that.

toerb:
I have changed it according to your suggestion.

FragmentedPacket:
Thanks for the contribution!

PR #175, opened 2020-04-12
Bugfix: Fix netbox_service to create new service with different protocols

### LINKED ISSUE #174: netbox_services still has problems..
##### ISSUE TYPE

 - Bug Report

##### STEPS TO REPRODUCE

create two identical services, except one is on TCP, the other should be using UDP.

The last service created wins.  The 1st service will be deleted.

The PR I submitted used 'protocol', you replaced it with 'port'.
##### EXPECTED RESULTS

I don't have services disappear based on protocol.

##### ACTUAL RESULTS

identical services with the only difference being protocol, the last service created will delete the other service.

FragmentedPacket:
Fixes #174

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -215,7 +215,7 @@ ALLOWED_QUERY_PARAMS = {
     "region": set(["slug"]),
     "rir": set(["slug"]),
     "role": set(["slug"]),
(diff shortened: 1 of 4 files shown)


FragmentedPacket:
@ThomasADavis Does this work as you need it to?

ThomasADavis:
LGTM

Issue #203, opened 2020-05-13
Inventory: Create groups based on rack groups, and rack roles


DouglasHeriot:
##### ISSUE TYPE

 - Feature Idea

###### Ansible:

v2.9.7

###### Netbox:

v2.8.3

###### Collection:

v0.2.0

##### SUMMARY

Currently you can specify for groups to be created by rack.  `/api/dcim/racks` is fetched, but nothing interesting is done with its data - there is no way to use existing `compose` option to create your own groups by a rack's group or role.

* `rack_role` should be a host var containing the slug of the rack's role
* `rack_groups` should be a host var with a list, similar to `regions` in that there may be parents. The ability to nest these was only introduced in Netbox v2.8.0 (netbox-community/netbox#1754)

When grouping by `rack_groups`, the `rack` group should be added as a child of `rack_group` group, to keep the group hierarchy clean.

PR #204, opened 2020-05-13
Inventory: Create groups based on rack groups, and rack roles

DouglasHeriot:
Implements #203 

Depends on pull request #202 - will rebase and mark as ready once that is merged.

Decided not to create a hierarchy of groups like I did for regions. Actually a little complicated to decide what groups to create when they're not all in use, or devices are filtered. Doesn't really impact usage or performance at all.


DouglasHeriot:
To fix tests, I need to work out the best way to ignore "parent rack group" missing from the inventory in Netbox 2.6-2.7, as only 2.8 supports nested rack groups.

DouglasHeriot:
Ok, I've fixed the tests now - this should be ready to review.

FragmentedPacket:
@DouglasHeriot Do you mind fixing the conflict?

I will also get this reviewed this weekend!

DouglasHeriot:
Yep, I'll try and sort this this weekend.

PR #212, opened 2020-05-23
Allow private key for validate_certs

### LINKED ISSUE #211: Accept string for validate_certs as done in pynetbox library
##### ISSUE TYPE

 - Feature Idea

###### Ansible:

v2.9
###### Netbox:

v2.7.6
###### Collection:

v0.2.1
##### SUMMARY

The pynetbox library contains a parameter ssl_verify which could be a boolean or a string.
Here is comment about this option in the pynetbox library:
```
:param bool/str,optional ssl_verify: Specify SSL verification behavior
        see: requests_.
```
String is useful to specify a folder containing the certs. With this capability it's possible to use self-signed certs.

Currently this project does not accept string for the parameter validate_certs. Is-it possible to change that ?

FragmentedPacket:
Fixes #211 

This will allow user to pass boolean or string pointing to a private key.

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -300,7 +300,7 @@ NETBOX_ARG_SPEC = dict(
     netbox_url=dict(type="str", required=True),
     netbox_token=dict(type="str", required=True, no_log=True),
     state=dict(required=False, default="present", choices=["present", "absent"]),
-    validate_certs=dict(type="bool", default=True),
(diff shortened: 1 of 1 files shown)


FragmentedPacket:
@smutel Do you mind testing this PR?

smutel:
Thank you.
I will test it Monday.

smutel:
I tested just right now and it is working as expected. Thanks.

PR #217, opened 2020-06-03
Added validate_certs option for lookup plugin

### LINKED ISSUE #216: nb_inventory plugin has no way to ignore SSL trust verification
##### ISSUE TYPE
 - Bug Report
 - Feature Idea

###### Ansible:
 2.9.9-1

###### Netbox:
2.7.3

###### Collection:
0.2.1

##### SUMMARY
The inventory query plugin has no way to disable SSL trust verification.  This is problematic because adding the trusted CA to AWX's docker containers doesn't work out of the box.  However this should be an option because the other modules in the netbox.netbox gallery appear to work properly

##### STEPS TO REPRODUCE

use loop: in a play and have something similar to:

- name: Print Tenant VLAN name
  vars:
    netbox_url_protocol: "https://{{netbox_url}}"
    vlan_name: {{ item.value.name }}
    tenant: "tenant"
  debug:
    msg: "The VLAN ID is : {{ item.value.vid }}"
    state: merged
  when: vlan_name == "my_vlan"
  loop: "{{ query('netbox.netbox.nb_lookup', 'vlans', 
                    api_endpoint=netbox_url_protocol,
                    ssl_verify=false,
                    api_filter='tenant=' + tenant,
                    token=netbox_token) }}"

```

##### EXPECTED RESULTS
Get the data back for this vlan

The VLAN ID is : 101

##### ACTUAL RESULTS
SSL ERROR in the play 
```
{
    "msg": "An unhandled exception occurred while running the lookup plugin 'netbox.netbox.nb_lookup'. Error was a <class 'requests.exceptions.SSLError'>, original message: HTTPSConnectionPool(host='netbox', port=443): Max retries exceeded with url: /api/ipam/vlans/?tenant=tntx&site=ashburn-va (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:897)'),))",
    "_ansible_no_log": null
}
```

FragmentedPacket:
Fixes #216 

@dakennedy

Do you mind testing with it set and without the new option set?

I just want to see if it sets the validate_certs to True by default so it should fail for you if it's not specified and if `false`, then work for you.

### DIFF
--- a/plugins/lookup/nb_lookup.py
+++ b/plugins/lookup/nb_lookup.py
@@ -49,6 +49,11 @@ DOCUMENTATION = """
                 - The API token created through Netbox
                 - This may not be required depending on the Netbox setup.
             required: False
+        validate_certs:
+            description:
+                - Whether or not to validate SSL of the NetBox instance
+            required: False
+            default: True
         key_file:
             description:
                 - The location of the private key tied to user account.
@@ -97,7 +102,7 @@ tasks:
 RETURN = """
   _list:
     description:
-      - list of composed dictonaries with key and value
+      - list of composed dictionaries with key and value
     type: list
 """
 
@@ -190,6 +195,7 @@ class LookupModule(LookupBase):
 
         netbox_api_token = kwargs.get("token")
         netbox_api_endpoint = kwargs.get("api_endpoint")
+        netbox_ssl_verify = kwargs.get("validate_certs")
         netbox_private_key_file = kwargs.get("key_file")
         netbox_api_filter = kwargs.get("api_filter")
         netbox_raw_return = kwargs.get("raw_data")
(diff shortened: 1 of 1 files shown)


dakennedy:
sure will give me a day and ill get back to you.  Since we are pulling from galaxy ill have to make some changes to import it in locally

FragmentedPacket:
@dakennedy No problem. Please keep me posted. I can always merge it, but prefer to make sure it works correctly before merging.

FragmentedPacket:
@dakennedy I will go ahead and merge this. I'm not entirely sure when I'll be cutting a new release, but hopefully soon.

If this breaks anything or doesn't work, just open a new issue and we can take another look, but the tests are passing and they use the lookup plugin so i'm fairly confident it isn't breaking anything.

PR #219, opened 2020-06-03
Cluster tenant

ThomasADavis:
This adds support to set the cluster tenant.

### DIFF
--- a/plugins/modules/netbox_cluster.py
+++ b/plugins/modules/netbox_cluster.py
@@ -162,7 +162,7 @@ def main():
                     cluster_type=dict(required=False, type="raw"),
                     cluster_group=dict(required=False, type="raw"),
                     site=dict(required=False, type="raw"),
-                    tenant=dict(required=False,type='str'),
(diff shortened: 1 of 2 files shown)


plugins/modules/netbox_cluster.py
```diff
@@ -159,6 +162,7 @@ def main():
                     cluster_type=dict(required=False, type="raw"),
                     cluster_group=dict(required=False, type="raw"),
                     site=dict(required=False, type="raw"),
+                    tenant=dict(required=False,type="str"),
                     comments=dict(required=False, type="str"),
```
FragmentedPacket, line 166: Could you add `tenant_group` here as well as that looks like an available field?

Also, could you add a test to the integration testing to create or update a cluster with a tenant?

FragmentedPacket:
If you run Black, I believe the tests should pass and then if you address the feedback, we can get this merged in.

FragmentedPacket:
Implementing in #230

PR #220, opened 2020-06-03
Inventory updates - fixes #187

### LINKED ISSUE #187: netbox_inventory_items problems..
##### ISSUE TYPE

 - Bug Report
 - Feature Idea

###### Ansible:

2.9.x

###### Netbox:

2.7.x

###### Collection:

0.20.0

##### SUMMARY

There is two issues/features.

1) All items are marked 'Manual entry'.  The discovered field is not being set/passed in.
simple line one fix - add 'discovered=dict(required=False, type='bool', default=True),' and it now marks the inventory properly.

2) Search/updates only use 1 field - name - which can be not very unique (ie, what I have more than one identical devices?).  The field size is also limited, so trying to cram a unique name in can be difficult. fix:  add serial, and description to the query.

I'll generate a PR when I'm done breaking things.  :)

ThomasADavis:
This adds 'serial' to make a unique search in the inventory using device, name, and serial

Also allows one to set the discovered flag.

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -198,7 +198,7 @@ ALLOWED_QUERY_PARAMS = {
     "device_type": set(["slug"]),
     "installed_device": set(["name"]),
     "interface": set(["name", "device", "virtual_machine"]),
-    "inventory_item": set(["name", "device","serial"]),
+    "inventory_item": set(["name", "device", "serial"]),
     "ip_address": set(["address", "vrf", "interface"]),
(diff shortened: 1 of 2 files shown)


FragmentedPacket:
This looks good for the most part, but do you mind updating the unit test that failed? Since `serial` was added, it just needs to be added to the expected since that is now an allowed query param for inventory_item.

FragmentedPacket:
@ThomasADavis I'm going to close this one in favor of #229.

I removed the serial option as it can mess with the searches, but #222 implements a way for users to define what options that are used to search for the objects.

Issue #15, opened 2019-09-21
Add A Few DCIM modules


FragmentedPacket:
##### ISSUE TYPE

 - Feature Idea

##### SUMMARY

Finish off most DCIM modules 
- regions
- device-bays
- inventory-items
- virtual-chassis (may need to update netbox_device to tie to virtual-chassis)

There are several DCIM remaining, but will let others contribute those if they have the use case:
- rack-reservations (low priority)
- console-port-templates (low priority)
- console-server-port-templates (low priority)
- power-port-templates (low priority)
- power-outlet-templates (low priority)
- interface-templates (low priority)
- front-port-templates (low priority)
- rear-port-templates (low priority)
- device-bay-templates (low priority)
- console-ports (low priority)
- console-server-ports (low priority)
- power-ports (low priority)
- power-outlets (low priority)
- front-ports (low priority)
- rear-ports (low priority)
- console-connections (low priority)
- power-connections (low priority)
- interface-connections (low priority)
- cables (low priority)
- power-panels (low priority)
- power-feeds (low priority)
- connected-device (low priority)

FragmentedPacket:
Implemented in #20

PR #235, opened 2020-06-18
Add modules for power management

toerb:
This PR will add the following modules:
- netbox_power_feed
- netbox_power_outlet
- netbox_power_outlet_template
- netbox_power_panel
- netbox_power_port
- netbox_power_port_template

All modules have a proper description and working integration tests.
They have already been mentioned in #15.


plugins/module_utils/netbox_utils.py
```diff
@@ -210,6 +226,12 @@
     "nat_inside": set(["vrf", "address"]),
     "parent_region": set(["slug"]),
     "platform": set(["slug"]),
+    "power_feed": set(["name"]),
```
FragmentedPacket, line 229: It appears that `power_panel` is requried in the `netbox_power_feed` module so I would probably add that as an `ALLOWED_QUERY_PARAM` for `power_feed` and any others for the other modules since you're requiring those.

plugins/modules/netbox_power_outlet.py
```diff
@@ -0,0 +1,319 @@
...
+        type: raw
+      name:
+        description:
+          - The name of the power outlet
+        required: true
+        type: str
+      type:
+        description:
+          - The type of the power outlet
+        choices:
```
FragmentedPacket, line 61: This may be a good option to just set to `raw` and not track the options as it may change, but totally up to you at this point. I love the documentation of this aspect, but the maintenance can be burdensome when new ones are added for `type`.

plugins/modules/netbox_power_outlet_template.py
```diff
@@ -0,0 +1,306 @@
...
+ANSIBLE_METADATA = {
+    "metadata_version": "1.1",
+    "status": ["preview"],
+    "supported_by": "community",
+}
+
+DOCUMENTATION = r"""
+---
+module: netbox_power_outlet_template
+short_description: Create, update or delete power outlets within Netbox
```
FragmentedPacket, line 20: `power outlet templates`?

plugins/modules/netbox_power_outlet_template.py
```diff
@@ -0,0 +1,306 @@
...
+description:
+  - Creates, updates or removes power outlets from Netbox
+notes:
+  - Tags should be defined as a YAML list
+  - This should be ran with connection C(local) and hosts C(localhost)
+author:
+  - Tobias Groß (@toerb)
+requirements:
+  - pynetbox
+version_added: '0.2.2'
```
FragmentedPacket, line 30: If you can get the changes in by Sunday, can you change version to `0.2.3`, but if you can't get the changes in by Sunday, change it to `0.2.4`

plugins/modules/netbox_power_port_template.py
```diff
@@ -0,0 +1,302 @@
...
+ANSIBLE_METADATA = {
+    "metadata_version": "1.1",
+    "status": ["preview"],
+    "supported_by": "community",
+}
+
+DOCUMENTATION = r"""
+---
+module: netbox_power_port_template
+short_description: Create, update or delete power ports within Netbox
```
FragmentedPacket, line 20: `power port templates`?

tests/integration/targets/latest/tasks/netbox_power_panel.yml
```diff
@@ -0,0 +1,85 @@
...
+    netbox_token: 0123456789abcdef0123456789abcdef01234567
+    data:
+      name: Power Panel
+      site: Test Site
+    state: present
+  register: test_two
+
+- name: "POWER_PANEL 2: ASSERT - Create duplicate"
+  assert:
+    that:
```
FragmentedPacket, line 42: Do you minding adding an update test to add to a rackgroup?

toerb: I totally forgot this test but interestingly it doesn't work anyway.
The `power-panel` endpoint is expecting a `rack_group` but the `rack_group` parameter is rewritten to `group` in `netbox_utils.py`.

Do you have any hint how I can disable the conversion from `rack_group` to `group` in this case?

plugins/modules/netbox_power_outlet_template.py
toerb: fixed to 0.2.3.

plugins/modules/netbox_power_outlet.py
toerb: I have seen both options in other modules and I decided to go with the explicit documentation. From my point of view this is nicer to read for users of this ansible collection.

plugins/module_utils/netbox_utils.py
toerb: Added, can you take a look if the new settings are right?

plugins/module_utils/netbox_utils.py
```diff
@@ -443,7 +470,9 @@ def _convert_identical_keys(self, data):
             if data.get("form_factor"):
                 temp_dict["type"] = data["form_factor"]
         for key in data:
-            if key in CONVERT_KEYS:
+            if self.endpoint == 'power_panels' and key == 'rack_group':
+                temp_dict[key] = data[key]
+            elif key in CONVERT_KEYS:
```
toerb, line 475: @FragmentedPacket: This workaround works but it is not very pretty.
I've found no other endpoint with a similar problem. It seems to be an inconsistency in the API from NetBox.

Do you have any suggestions on how to proceed regarding this?

tests/integration/targets/latest/tasks/netbox_power_panel.yml
FragmentedPacket: It looks like your update resolved this. I would have handled it the same way at this point, but I'll look further into it.

plugins/module_utils/netbox_utils.py
FragmentedPacket: I think this is the way I would have done it at the moment anyways. We may want to update the unit tests to account for this.

plugins/module_utils/netbox_utils.py
FragmentedPacket: They look correct to me. I will go ahead and merge this PR.

Thank you so much for these PRs you have submitted! They were really high quality and I appreciate that!

I think the docs may be a bit out of date, but were they helpful?
Was it easy to make these new modules?
Would you change anything about the docs or contribution process?

Looking for honest feedback so don't be scared to be honest :)

toerb: > Thank you so much for these PRs you have submitted! They were really high quality and I appreciate that!

Thanks for the kind words! :)
I am going to add a few more modules next week.

> I think the docs may be a bit out of date, but were they helpful?

They were really helpful! Without the docs it would have been much more complicated.

> Was it easy to make these new modules?

The first one was a bit tricky. For the modules afterwards the biggest problem is doing clean work and not being careless.

> Would you change anything about the docs or contribution process?
While the documentation is already very good I would probably structure it a little bit different.
I will have a look into it next week, after finishing the other modules.
The biggest pain points are probably the unorderly structure (at least for my think process) and implicit knowledge which is not available to beginners.

PR #236, opened 2020-06-19
Add modules for console management

toerb:
This PR will add the following modules:
- netbox_console_port
- netbox_console_port_template
- netbox_console_server_port
- netbox_console_server_port_template

All modules have a proper description and working integration tests.
They have already been mentioned in #15.


plugins/modules/netbox_console_port.py
```diff
@@ -0,0 +1,219 @@
...
+      - The token created within Netbox to authorize API access
+    required: true
+    type: str
+  data:
+    type: dict
+    required: true
+    description:
+      - Defines the console port configuration
+    suboptions:
+      device:
```
FragmentedPacket, line 48: Same as the other PR, may want to add this to the `ALLOWED_QUERY_PARAMS` since it's required.

plugins/modules/netbox_console_port_template.py
```diff
@@ -0,0 +1,206 @@
...
+description:
+  - Creates, updates or removes console port templates from Netbox
+notes:
+  - Tags should be defined as a YAML list
+  - This should be ran with connection C(local) and hosts C(localhost)
+author:
+  - Tobias Groß (@toerb)
+requirements:
+  - pynetbox
+version_added: '0.2.2'
```
FragmentedPacket, line 30: Same comment as the other PR. Change to `0.2.3` if you can get the fixes in by Sunday or `0.2.4` if not.

FragmentedPacket:
Do you mind resolving the conflict and then committing it? I will get it merged after that. I think we may have to do that for the other PR as well after this one is merged.

Issue #137, opened 2020-03-16
New Module Netbox Virtual Chassis


nathbooth:
ISSUE TYPE

    Feature Idea

SUMMARY

 -  virtual-chassis

nathbooth:
Currently WIP, will raise PR when complete

toerb:
@FragmentedPacket: I am currently trying implementing this module but I got stuck.

virtual-chassis has one required field, called `master`.
But regarding the documentation it is not filterable by this field.
I don't see a good way to add this module without the option to filter after `master` or `master_id`.

I already tried by using the `q` filter but I did not succeed so far and it bloats up the code.
Do you have an idea on how to implement this modules correctly?

FragmentedPacket:
Are your questions still valid after submitting the PR?

toerb:
The PR is now using the method to filter by `q`. While this is not pretty I think it's currently the best way to filter for virtual-chassis in the NetBox API.

FragmentedPacket:
Ok awesome. I will take the time and review it today. Thanks for your contributions!

toerb:
@nathbooth: This feature request is implemented now. It is shipped in v0.3.0.

PR #251, opened 2020-06-30
Add modules for cable, device_bay_template and virtual_chassis management

toerb:
As referenced in #137 this is the prototype for handling virtual chassis.

It is working but not very pretty in my opinion.
The main reason for this is the missing option to filter for `master` or `master_id` in the `virtual-chassis` endpoint.
This solution uses the `q` filter to search for already existing virtual-chassis objects.


toerb:
@FragmentedPacket: Any comments so far?

toerb:
@FragmentedPacket: I have added modules for cable management and device_bay_template to this pull request since I don't want to rebase the code.

Please have a look at the cable implementation. I think I found a quite good solution for now.

FragmentedPacket:
I will spend some time this weekend and review everything!

Issue #248, opened 2020-06-28
Update modules to work with pynetbox >5.0.0


FragmentedPacket:
##### ISSUE TYPE

 - Bug Report

Need to change the way the `self.nb` object is created due to changes in Pynetbox 5.0.0

jvanderaa:
Are we getting a bunch of benefit out of using pynetbox? Would it be worth looking at moving to just using requests, preventing the need for urgent refactoring? Since it looks like the new method requires a pre-built session instead of establishing one within the SDK.

Just a question, not sure of what it would look like at the moment. I have some work where I do need to do something similar, but probably on a small scale.

FragmentedPacket:
It would take a lot of work to refactor the code and tests for this so not sure it's worth it at this point, but possibly sometime in the future to remove the dependency on an external SDK would probably be ideal, although, it is nice to have some of the built-in capabilities that come with `pynetbox`.

PR #263, opened 2020-07-08
Update module to work with pynetbox 5.0

smutel:
Fix issue #248


FragmentedPacket:
In the `.travis.yml` we hard-coded the `pynetbox` to `4.3.3` to work around this issue. Do you mind removing the pinned version and just let it install the latest and I think the tests will hopefully pass.

If that doesn't fix it, I will have to do some research around the testing error.

FragmentedPacket:
It may also require the import of requests to be right below the attempt of importing `pynetbox` after looking into this a bit further since it's a 3rd party package.

smutel:
I add a pip list in https://travis-ci.com/github/netbox-community/ansible_modules/jobs/359616773 and we can see that request is correctly installed. I don't understand why the ci is not passing. I will continue to investigate ...

FragmentedPacket:
I may take this over but I think if you move the import of request into where pynetbox is, some of the issues will be fixed

FragmentedPacket:
I've fixed some of the stuff, but I think in the 5.0.0 release, it changed how it returned the available_ips and prefixes so have to fix that as well.

I will most likely close this one out once I have it fixed, but make sure to credit you.

I also appreciate your effort on this and the communication on both this project and `pynetbox`, it is greatly appreciated!

FragmentedPacket:
Made a few updates with other changes involved with pynetbox >=5.0.0 on #269 

Thanks a lot for your work @smutel !

PR #277, opened 2020-07-14
Make a copy of ipaddress before modifying to not break caching.

### LINKED ISSUE #276: Inventory: IPaddresses of interfaces is lost when using cache.
##### ISSUE TYPE

 - Bug Report

###### Ansible:

ansible 2.9.10

###### Netbox:

v2.8.7

###### Collection:

v0.3.0

##### SUMMARY

When using caching together with nb_inventory and expanding interfaces IP-address mapping gets lost when using cache.

##### STEPS TO REPRODUCE

The issue is reproducible by configure nb_inventory and using cache.

```yaml
cache: true
cache_plugin: jsonfile
cache_timeout: 3600
api_endpoint: https://netbox-lab.netnod.se
interfaces: true
```

Take any device with a interface that have ip-address assigned to it and run the following command:
```shell
$ rm /tmp/ansible/ansible_inventory_netbox*
$ ansible-inventory --host ag1b-bck.gbg.netnod.se 2>/dev/null | jq '.interfaces[] | select(.name == "Ethernet1").ip_addresses[].address'
"172.20.9.2/31"
$ ansible-inventory --host ag1b-bck.gbg.netnod.se 2>/dev/null | jq '.interfaces[] | select(.name == "Ethernet1").ip_addresses[].address'
$ 
```

I have done some debugging and it looks like the issue is due to the ip-addresses query results from netbox is modified and the modified data is written to the cache. 

Removal of interface mapping is done here:
(https://github.com/netbox-community/ansible_modules/blob/devel/plugins/inventory/nb_inventory.py#L837) 

The ip-address-> interface mapping is needed to attach the ip-addresses to the interfaces. 
And in consecutive runs this is not available.

##### ACTUAL RESULTS

```

```

matte4711:
Fixes #276

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -831,10 +831,13 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             interface_id = ipaddress["interface"]["id"]
             ip_id = ipaddress["id"]
 
-            self.ipaddresses_lookup[interface_id][ip_id] = ipaddress
+            # We need to copy the ipaddress entry to preserve the original in case caching is used.
+            ipaddress_copy = ipaddress.copy()
+
+            self.ipaddresses_lookup[interface_id][ip_id] = ipaddress_copy
 
             # Remove "interface" attribute, as that's redundant when ipaddress is added to an interface
-            del ipaddress["interface"]
+            del ipaddress_copy["interface"]
 
     @property
     def lookup_processes(self):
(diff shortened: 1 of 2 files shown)


FragmentedPacket:
I'm going to approve this and then implement proper tests for this.

We should enable caching (which you have done), but also test that cache file against an expected cache file, but I want to overhaul some of this to allow contributors to run this stuff locally.

I opened issue #305 to track that though.

Thanks for your contribution!

PR #286, opened 2020-07-22
Make group_by work with services

ptoal:
The group_by has apparent support for grouping by "services", but the functionality doesn't work, as it provides a list of dicts, which are not accepted.  This PR makes group_by accept 'service', or 'services' (depending on plurality), and adds hosts to groups based on the service.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -127,6 +127,7 @@ DOCUMENTATION = """
                 - cluster_type
                 - cluster_group
                 - is_virtual
+                - services
             default: []
         group_names_raw:
             description: Will not add the group_by choice name to the group names
@@ -1054,6 +1055,11 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             else:
                 # Don't create the inverse group
                 return None
+
+        # Special case. Extract name from service, which is a hash.
+        if grouping == "services":
+            group = group["name"]
+            grouping = "service"
 
         if self.group_names_raw:
             return group


FragmentedPacket:
Do you mind running `black .` and then fixing the tests that failed?

ptoal:
> Do you mind running `black .` and then fixing the tests that failed?

Sorry, but I tried to get the environment set up on my Mac to run the tests, but after spending an hour trying to figure it out, I had to give up.  The [documentation](https://github.com/netbox-community/ansible_modules/blob/devel/CONTRIBUTING.md) for this doesn't seem correct, but maybe it's just my environment.

ptoal:
So... I spent a bit more time on this today, and just can't figure out how to get the tests to run properly.  The contributing doc suggests running pytest -vv, which blows up immediately for me, and trying to manually execute the steps in the travis file gets me as far as `ansible-test integration -v --coverage --python $PYTHON_VER inventory` before I get an error:
```
+ ./compare_inventory_json.py ./files/test-inventory.json /private/tmp/travis/.ansible/collections/ansible_collections/netbox/netbox/tests/output/.tmp/output_dir/test-inventory.json
+ exit 1
NOTICE: To resume at this test target, use the option: --start-at inventory
ERROR: Command "./runme.sh -v" returned exit status 1.
```
I think I'm going to need a little help understanding how the tests are supposed to work, and get a working test run before I can actually contribute with tests that pass.

FragmentedPacket:
Yeah sorry about the outdated documentation. I'm going to be working on the docs and get them hosted on RTD and then containerize the tests since we changed the tests significantly and it would be easiest to run the tests locally within containers.

If you run `black .` within the parent directory, it should fix the black failure.

The other test failing pertains to [this file](https://github.com/netbox-community/ansible_modules/blob/devel/tests/integration/targets/inventory/files/test-inventory-options.json) and the test shows what it needs to look like. Although, we should make sure it is returning the expected output when not pluralizing the services. The other file appears to be the `test-inventory.json` file

plugins/inventory/nb_inventory.py
```diff
@@ -127,6 +127,8 @@
                 - cluster_type
                 - cluster_group
                 - is_virtual
+                - service
```
DouglasHeriot, line 130: "service" should never be an option - it's always a list.

plugins/inventory/nb_inventory.py
```diff
@@ -387,6 +389,7 @@ def _pluralize_group_by(self, group_by):
             "platform": "platforms",
             "device_type": "device_types",
             "manufacturer": "manufacturers",
+            "service": "services",
```
DouglasHeriot, line 392: No more properties should ever be added here - it's for legacy things like "sites" that didn't make any sense, because "site" is always singular (a NetBox device will only ever have a single site, and I do not forsee that API ever changing)

DouglasHeriot:
Actually, on a closer look at the diff in the tests, I have to disagree with this implementation.

Grouping by services does sound useful, but the full service objects should not be removed from the output.

A use case example - I'm currently planning to use tags and custom fields on services to generate load-balancer configurations.

```
            "Test Nexus One": {
                "$delete": {
                    "services": [
                        {
                            "custom_fields": {},
                            "description": "",
                            "device": {
                                "display_name": "Test Nexus One",
                                "id": 4,
                                "name": "Test Nexus One"
                            },
                            "id": 3,
                            "ipaddresses": [],
                            "name": "telnet",
                            "port": 23,
                            "virtual_machine": null
                        }
                    ]
                },
                "$insert": {
                    "service": [
                        "telnet"
                    ]
                }
            },

```

ptoal:
It took me a bit to understand what you were saying, but I think I get it now.  My initial confusion was that I presumed that group_extractors was used only for group_by, and I concluded that the services extractor wasn't properly implemented.

What I'm now understanding is that group_extractors isn't used just to provide a list of groups that a host should be in, but it is overloaded to provide host variables as well.  Is that right?

ptoal:
Alright.  With a bit more understanding, I took another shot at this.  I haven't written any tests, yet, but the existing ones pass, now.

FragmentedPacket:
Could you add this option to one of the tests? We want to make sure it makes the correct groups here.

Then I think we can merge it.

FragmentedPacket:
Closing this in favor of #310 to get testing added and get this change merged in.

@ptoal Thanks for your work on this!

ptoal:
Thanks for carrying it over the finish line. :)

FragmentedPacket:
Np. Thanks for the work!

I've neglected the documentation for this stuff and really need to revamp it so it's easier on you guys to be able to contribute.

PR #295, opened 2020-08-04
NetBox Device Interface - type is a required argument

sc68cal:
The documentation says it is not, but the code states it is a required parameter

### DIFF
--- a/plugins/modules/netbox_device_interface.py
+++ b/plugins/modules/netbox_device_interface.py
@@ -67,7 +67,7 @@ options:
             Form factor of the interface:
             ex. 1000Base-T (1GE), Virtual, 10GBASE-T (10GE)
             This has to be specified exactly as what is found within UI
-        required: false
+        required: true
         type: str
       enabled:
         description:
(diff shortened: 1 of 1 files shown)


FragmentedPacket:
@sc68cal I thought I had commented on this so sorry for the delay.

It's not required when deleting as we're just looking for a few items to find the unique device, but it might be worth it to add `type` to the `required_if` and the `state` being present. I think that will help with providing an error as well if you want to change the PR to that.

ThomasADavis:
It's also not required when your adding/removing lags; ie:

```      netbox_device_interface:
        netbox_url: "{{ netbox_url }}"
        netbox_token: "{{ netbox_token }}"
        data:
          device: "{{ item[1] }}"
          name: "swp{{ item[0].uplink_port }}"
          mtu: 9216
          lag:
              name: "bond{{ item[0].uplink_port}}"
        state: present
```

works fine.

FragmentedPacket:
@ThomasADavis That's a great point. The only time some of the parameters matter is on creation, but when updating or deleting, they aren't needed.

I guess we can really cancel this PR since my latest proposed thing will break what @ThomasADavis brought up.

I just made a new doc site so I should probably add some examples of using the modules with the different options.

I will open a new issue to document these. Thanks for your time and contribution btw!

PR #296, opened 2020-08-05
Update README.md

sy-na-github:
Updated the section on nb_lookup.py because it does not play well with MacOS.  Some certain arguments and context are added so that others do not have to go through the same tshooting steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ Using the **collections** lookup plugin at the task level
       vars:
         query_result: "{{ query('netbox.netbox.nb_lookup', 'sites', api_endpoint='http://localhost:32768', token='0123456789abcdef0123456789abcdef01234567') }}"
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -201,6 +201,7 @@ Using the **collections** lookup plugin at the task level
       vars:
         query_result: "{{ query('netbox.netbox.nb_lookup', 'sites', api_endpoint='http://localhost:32768', token='0123456789abcdef0123456789abcdef01234567') }}"
 ```
+Note that if you are on MacOS and are running into ```ERROR! A worker was found in a dead state``` errors, try running the playbook with a ```env no_proxy='*'``` tag in front of the playbook.  This is a known issue with MacOS as per this reference: https://github.com/ansible/ansible/issues/32554#issuecomment-642896861
```
FragmentedPacket, line 204: Do you mind changing the plus to `>`

PR #323, opened 2020-08-27
feat: add follow_redirects option

### LINKED ISSUE #318: Request to netbox does not follow redirects on 308
##### ISSUE TYPE

 - Bug Report

##### SUMMARY

Requests does not follow redirects.

##### STEPS TO REPRODUCE

- Configure netbox to use HTTP
- Add nginx ingress controller in front of it with HTTPS and HTTP to HTTPS redirect (default to 308)
- Query netbox and get pagination links

##### EXPECTED RESULTS
```log
Fetching: https://netbox.corp/api/dcim/devices/?limit=0
Fetching: https://netbox.corp/api/dcim/devices/?limit=1000&offset=1000
```

##### ACTUAL RESULTS

```log
Fetching: https://netbox.corp/api/dcim/devices/?limit=0
Fetching: http://netbox.corp/api/dcim/devices/?limit=1000&offset=1000
...
File "/usr/lib64/python3.6/urllib/request.py", line 673, in redirect_request
    raise HTTPError(req.full_url, code, msg, headers, fp)
urllib.error.HTTPError: HTTP Error 308: Permanent Redirect
```

Digging a bit more into it, it looks like `open_url` ansible function follows redirects by default.
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py#L1130
https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/urls.py#L689
https://github.com/netbox-community/ansible_modules/blob/devel/plugins/inventory/nb_inventory.py#L262

But, as netbox sends back a 308 redirect and urllib has no handler for it, it sends back an error: https://docs.python.org/3/library/urllib.request.html#httpredirecthandler-objects
> Otherwise, raise HTTPError if no other handler should try to handle this URL

A quick fix would be to add `follow_redirect=True` to the `open_url` call.

aslafy-z:
Fixes #318

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -264,6 +264,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                     headers=self.headers,
                     timeout=self.timeout,
                     validate_certs=self.validate_certs,
+                    follow_redirect=True,
                 )
             except urllib_error.HTTPError as e:
                 """This will return the response body when we encounter an error.
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -35,6 +35,12 @@ DOCUMENTATION = """
                 - Allows connection when SSL certificates are not valid. Set to C(false) when certificates are not trusted.
             default: True
             type: boolean
+        follow_redirects:
+            description:
+                - Determine how redirects are followed.
+                - By default, I(follow_redirects) is set to uses urllib2 default behavior.
+            default: urllib2
+            choices: ['urllib2', 'all', 'yes', 'safe', 'none']
         config_context:
             description:
                 - If True, it adds config_context in host vars.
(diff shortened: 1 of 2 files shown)


plugins/module_utils/netbox_utils.py
```diff
@@ -501,7 +501,10 @@ def _validate_query_params(self, query_params):
         # Fetch the OpenAPI spec to perform validation against
         base_url = self.nb.base_url
         junk, endpoint_url = nb_endpoint.url.split(base_url)
-        response = open_url(base_url + "/docs/?format=openapi")
```
FragmentedPacket, line 504: Do you mind reverting this change? We don't use this method currently and `pynetbox` has a method we can use to do exactly what this function is doing.

plugins/inventory/nb_inventory.py
```diff
@@ -264,6 +264,7 @@ def _fetch_information(self, url):
                     headers=self.headers,
                     timeout=self.timeout,
                     validate_certs=self.validate_certs,
+                    follow_redirect=True,
```
FragmentedPacket, line 267: Do you mind changing this to an option?

1. Update documentation to add `follow_redirects` option
2. Gather option within the `parse` function that we grab all the possible options
3. Change L267 to use `self.follow_redirects`

aslafy-z: Just implemented it, keeping `urllib2` as the default value to avoid introducing a breaking change.

FragmentedPacket (approved):

FragmentedPacket:
Even though pipeline is failing, going to merge.

The reason for the failure is related to: https://github.com/ansible/ansible/issues/71602

PR #334, opened 2020-09-09
IP addresses missing from virtual machine interfaces

### LINKED ISSUE #321: Inventory - IP addresses missing from virtual machine interfaces
##### ISSUE TYPE

 - Bug Report

###### Ansible:

ansible 2.9.11
###### Netbox:

2.9.1
###### Collection:

1.0.1
##### SUMMARY

When interfaces is set to TRUE the IP address array is empty for all VMs
##### STEPS TO REPRODUCE

set `interfaces: True` in inventory yml

```yaml
api_endpoint: http://netbox
validate_certs: True
config_context: False
services: True
interfaces: True
fetch_all: False
plurals: False

group_by:
  - role

query_filters:
  - cluster_id: 1

compose:
  netbox_vm_id: id
```

##### EXPECTED RESULTS

```json
interfaces": [
  {
   "description": "",
   "enabled": true,
   "id": 4,
   "ip_addresses": [
       "List of IP objects": "Example"
    ],
   "mac_address": null,
   "mode": null,
   "mtu": null,
   "name": "ens18",
   "tagged_vlans": [],
   "tags": [],
   "untagged_vlan": null,
   "url": "http://netbox/api/virtualization/interfaces/4/",
   "virtual_machine": {
    "id": 13,
    "name": "VM NAME",
    "url": "http://i/api/virtualization/virtual-machines/13/"
   }
  }
 ]
```
##### ACTUAL RESULTS

```json
interfaces": [
  {
   "description": "",
   "enabled": true,
   "id": 4,
   "ip_addresses": [],
   "mac_address": null,
   "mode": null,
   "mtu": null,
   "name": "ens18",
   "tagged_vlans": [],
   "tags": [],
   "untagged_vlan": null,
   "url": "http://netbox/api/virtualization/interfaces/4/",
   "virtual_machine": {
    "id": 13,
    "name": "VM NAME",
    "url": "http://i/api/virtualization/virtual-machines/13/"
   }
  }
 ]
```

Might be related to this.
netbox-community/netbox#4721
https://github.com/netbox-community/ansible_modules/issues/319

haxorof:
Fixes #321

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -599,11 +599,19 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
             interfaces = list(interfaces_lookup[host["id"]].values())
 
+            before_netbox_v29 = bool(self.ipaddresses_lookup)
             # Attach IP Addresses to their interface
             for interface in interfaces:
-                interface["ip_addresses"] = list(
-                    self.ipaddresses_lookup[interface["id"]].values()
-                )
+                if before_netbox_v29:
+                    interface["ip_addresses"] = list(
+                        self.ipaddresses_lookup[interface["id"]].values()
+                    )
+                else:
+                    interface["ip_addresses"] = list(
+                        self.vm_ipaddresses_lookup[interface["id"]].values()
+                        if host["is_virtual"]
+                        else self.device_ipaddresses_lookup[interface["id"]].values()
+                    )
 
             return interfaces
         except Exception:
@@ -907,6 +915,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         # Construct a dictionary of lists, to allow looking up ip addresses by interface id
         # Note that interface ids share the same namespace for both devices and vms so this is a single dictionary
         self.ipaddresses_lookup = defaultdict(dict)
+        # NetBox v2.9 and onwards
+        self.vm_ipaddresses_lookup = defaultdict(dict)
+        self.device_ipaddresses_lookup = defaultdict(dict)
 
         for ipaddress in ipaddresses:
             # As of NetBox v2.9 "assigned_object_x" replaces "interface"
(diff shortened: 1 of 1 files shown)


haxorof:
Fix not correct see referenced issue for comment. Will look into this.

haxorof:
Did a correction but happy to get some feedback on that. Feel that there might be some nicer way of doing it. 😄

ThomasADavis:
If I remember correctly, there is a pynetbox function to return the netbox API version, which should be safer than what you tried to do; ie:

```>>> import pynetbox
>>> nb = pynetbox.api(
...     'http://localhost:8000',
...     private_key_file='/path/to/private-key.pem',
...     token='d6f4e314a5b5fefd164995169f28ae32d987704f'
... )
>>> nb.version
'2.6'
>>>```

this is really the correct way to version things.

**Comment by haxorof** [comment]
@ThomasADavis Currently pynetbox is not used in the nb_inventory.py what I can see so you suggestion is to introduce that here? Feels that it can be a bit risky to introduce that dependency now.

**Comment by ThomasADavis** [comment]
A function to get the version of the netbox that nb_inventory is talking to should be added, for it's also probably going break on 'tags' also.

The netbox version is returned on every rest call as part of the header information; see https://github.com/netbox-community/netbox/issues/1223

@FragmentedPacket it's up to you...  the inventory use direct REST api calls.

**Comment by FragmentedPacket** [comment]
Keeping it via the API calls is fine. I've been contemplating whether it's worth it to move away from `pynetbox` with some of the issues that have arisen over use of it in this project.

I'll take a look at this PR soon and provide any feedback once I've had time to actually take a good look at it.

Thanks for your work on this though!

File: plugins/inventory/nb_inventory.py [hunk 1]
```diff
@@ -599,11 +599,19 @@ def extract_interfaces(self, host):
 
             interfaces = list(interfaces_lookup[host["id"]].values())
 
+            before_netbox_v29 = bool(self.ipaddresses_lookup)
             # Attach IP Addresses to their interface
             for interface in interfaces:
-                interface["ip_addresses"] = list(
-                    self.ipaddresses_lookup[interface["id"]].values()
-                )
+                if before_netbox_v29:
```
FragmentedPacket, line 605: Do we even need to to determine if it's < 2.9?

Could we just have the two lookups, device and VM ip address lookups? So we handle the difference in versions in the code below by determining which bucket to throw them into?

FragmentedPacket (changes requested):

plugins/inventory/nb_inventory.py
haxorof: If you look in the changes I did in `refresh_ipaddresses` then you see that for NetBox v2.9 and later I will know if it is an IP address is connected to a VM or device by looking at the `assigned_object_type`. But in earlier NetBox versions this information is not available like that to determine which bucket to throw them into. Or what kind of change do you suggest? 

It would have been great if I just could have the the two lookups and skip the `ipaddresses_lookup` completely but as said I do not see that I have the information available in earlier version to put IPs in the two buckets.

FragmentedPacket: Ok I do see what you're saying now. And we can't act off anything else with the way the inventory is setup to differentiate.

I will approve and merge this in.

FragmentedPacket (approved):

PR #349, opened 2020-10-02
Adding support for netbox plugins in ansible lookup plugin

fach:
The ansible lookup plugin doesn't appear to support netbox plugins for querying model extensions. This PR adds said functionality. 

One caveat (and looking for suggestions): there doesn't seem to be a way to derive whether a plugin or the associated term/model exist upstream until request time. In the event a plugin or model don't exist, the user will receive the dreaded `RequestError` from pynetbox:
```
An unhandled exception occurred while running the lookup plugin 'nb_lookup'. Error was
a <class 'pynetbox.core.query.RequestError'>, original message: The requested url: https://
localhost/api/plugins/mycustomstuff/bgp-sessions/?device__name=R1-Device could not be found.
```
This doesn't seem terrible but it would be nice to catch this when deriving the correct endpoint, as is done for the core terms/models.

### DIFF
--- a/plugins/lookup/nb_lookup.py
+++ b/plugins/lookup/nb_lookup.py
@@ -11,6 +11,7 @@ A lookup function designed to return data from the Netbox application
 
 from __future__ import absolute_import, division, print_function
 
+import functools
 from pprint import pformat
 
 from ansible.errors import AnsibleError
@@ -44,6 +45,10 @@ DOCUMENTATION = """
         api_filter:
             description:
                 - The api_filter to use.
+            required: False
+        plugin:
+            description:
+                - The Netbox plugin to query
             required: False
         token:
             description:
@@ -98,6 +103,17 @@ tasks:
   - name: "Obtain secrets for R1-Device"
     debug:
       msg: "{{ query('nb_lookup', 'secrets', api_filter='device=R1-Device', api_endpoint='http://localhost/', token='<redacted>', key_file='~/.ssh/id_rsa') }}"
+
+# Fetch bgp sessions for R1-device
+tasks:
+  - name: "Obtain bgp sessions for R1-Device"
+    debug:
+      msg: "{{ query('nb_lookup', 'bgp_sessions',
+                     api_filter='device=R1-Device',
+                     api_endpoint='http://localhost/',
+                     token='<redacted>',
+                     plugin='mycustomstuff') }}"
+
 """
 
 RETURN = """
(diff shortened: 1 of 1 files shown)


fach:
Hm, I don't think the build failure is due to this PR.

FragmentedPacket:
Most likely not. I will review this soon and I'm working on getting the repo working with NB 2.9

plugins/lookup/nb_lookup.py
```diff
@@ -222,11 +254,13 @@ def run(self, terms, variables=None, **kwargs):
 
         results = []
         for term in terms:
-
-            try:
-                endpoint = get_endpoint(netbox, term)
-            except KeyError:
-                raise AnsibleError("Unrecognised term %s. Check documentation" % term)
+            if netbox_plugin:
```
FragmentedPacket, line 257: I believe we should be doing a try/except block here and then return an error if we hit an `AttributeError`.

Return something like: `AnsibleError("%s does not exist in %s plugin." % (term, plugin)`

This is because `get_plugin_endpoint` uses the `getattr` which can return an `AttributeError`

FragmentedPacket (changes requested):

plugins/lookup/nb_lookup.py
fach: Checking for an `AttributeError` doesn't work here, unfortunately. Because these plugin models aren't build into the core of pynetbox, the endpoint seems to be merely a string used to craft the api url and only eval'd at request time. Any plugin/model name appears to be valid, regardless of whether the netbox instance actually has it.

There is an `installed_plugins()` call that can be made to at least check for a valid plugin name but I don't see an option for verifying the model itself:
```
[{'name': 'Test', 'package': 'netbox_test', 'author': 'test', 'author_email': 'test@test.com', 'description': 'test plugin', 'verison': '0.1'}]

FragmentedPacket: Ah yes. Thanks for the explanation of that.

I'm thinking we may want to wrap the `filter` and `all` in a try/except block looking for a `RequestError` and then maybe inspect a bit further for a 404 error and then provide a meaningful error message such as; `Plugin endpoint is not valid. Please validate the provided plugin has the requested endpoint. {url}` or something like that.

FragmentedPacket:
@fach Thanks for your contribution.

I took over this change in #360 just to refactor a bit, but leaving your commit to make sure you get credited!

Thanks for your contribution!

PR #350, opened 2020-10-03
Update netbox ansible-modules to support Netbox 2.9

mrevilme:
This is my go at trying to get netbox-ansible-modules to support Netbox 2.9 style of ipaddress assignments towards both dcim.interface and virtualization.vminterface.

Doing this PR to get the ball rolling.


FragmentedPacket:
@mrevilme Thanks for this PR! The fix is actually in #345, but need to fix all the testing that broke between Ansible 2.10 and NetBox 2.9.

Thanks for your work! It has some similarities between the two which is awesome to we were heading down the same road.

Issue #352, opened 2020-10-05
nb_inventory: add status to host_vars


1977er:
##### SUMMARY
It would be nice to have the Netbox (device/vm) status field variable available as host_vars. 

##### STEPS TO REPRODUCE
```
ansible-inventory -i netbox.yml --host my_test_host
```

##### EXPECTED RESULTS
```
{
...
 "status": {
    "label": "Decommissioning",
    "value": "decommissioning"
 }
...
}
```

##### ACTUAL RESULTS
none yet

FragmentedPacket:
Fixed by #353

PR #353, opened 2020-10-05
Enhancement: nb_inventory, add status (implements #352)

1977er:
My first PR.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -365,6 +365,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             "disk": self.extract_disk,
             "memory": self.extract_memory,
             "vcpus": self.extract_vcpus,
+            "status": self.extract_status,
             "config_context": self.extract_config_context,
             "local_context_data": self.extract_local_context_data,
             "custom_fields": self.extract_custom_fields,
@@ -449,6 +450,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
     def extract_vcpus(self, host):
         return host.get("vcpus")
+
+    def extract_status(self, host):
+        return host.get("status")
 
     def extract_memory(self, host):
         return host.get("memory")


plugins/inventory/nb_inventory.py
```diff
@@ -450,6 +451,9 @@ def extract_disk(self, host):
     def extract_vcpus(self, host):
         return host.get("vcpus")
 
+    def extract_status(self, host):
+        return host.get("status")
```
FragmentedPacket, line 455: Should we return either the label or value from the `status` dictionary?

plugins/inventory/nb_inventory.py
```diff
@@ -365,6 +365,7 @@ def group_extractors(self):
             "disk": self.extract_disk,
```
FragmentedPacket, line 365: Do you mind updating the inventory related tests to add the status for each device after you fix the one question below?

The inventory tests are shown in the failed tests. I should be done with fixing the tests tonight as well if you want to rebase off devel once I let you know and then just fix the failing inventory tests from there.

FragmentedPacket (changes requested):

plugins/inventory/nb_inventory.py
1977er: Why limit ourselves? In doubt I'd prefer "value".

plugins/inventory/nb_inventory.py
1977er: Sorry, I'm a newbie. I don't know, how to work with this test framework. :-(

metabsd:
You can use the compose option if you want to extract status.

plugins/inventory/nb_inventory.py
FragmentedPacket: This can be resolved by @metabsd's comment on the PR.

plugins/inventory/nb_inventory.py
FragmentedPacket: You'll need to look at what failed in the Travis CI run and then correct the corresponding files for the inventory tests. The tests will tell you what each file is expecting.

Those are all stored in `tests/integration/targets/inventory/files`

FragmentedPacket:
@1977er Closing this for #359, but keeping your commits to provide you credit!

I appreciate your PR and wanted to make sure testing was updated as well.

Cheers!

Issue #135, opened 2020-03-12
New Module to support config_contexts


smolz:
##### ISSUE TYPE

 - Feature Idea

###### Ansible:

2.9
###### Netbox:

2.7.8
###### Collection:

v0.1.10
##### SUMMARY

There is no module to support the creation of config context endpoints, in the API `extras/config_context`
##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

PR #364, opened 2020-10-13
New module: netbox_config_context (#135)

pkorovin:
New module: netbox_config_context (#135)


plugins/module_utils/netbox_extras.py
```diff
@@ -34,9 +39,80 @@ def run(self):
...
+            tenant_groups = data["tenant_groups"]
+            nb_app = getattr(self.nb, "tenancy")
+            nb_tenant_group = getattr(nb_app, "tenant_groups")
+            tenant_groupids = []
+            for tenant_group in tenant_groups:
+                tenant_groupids.append(nb_tenant_group.get(slug=(tenant_group)).id)
+                data["tenant_groups"] = tenant_groupids
 
-        data["slug"] = self._to_slug(name)
+        if data.get("tenants"):
```
FragmentedPacket, line 108: I think we should most these to resolving via the normal methods. This would require adding the plurals to the `CONVERT_TO_ID` and then `ALLOWED_QUERY_PARAMS` for each.

This should allow people to specify them either via string or dictionary to specify extra context to find the specific endpoint.

FragmentedPacket (changes requested):
The same as the `tags` module, can you please add integration tests as well?

plugins/module_utils/netbox_extras.py
pkorovin: @FragmentedPacket Mikhail, integration test were added, all assertions passed for netbox 2.8/2.9. As of resolving, could you merge it for now as it is, and I'll try to modify the code as suggested?

FragmentedPacket: I'd rather not merge something that still needs work. I can take this over and get it across the finish line since I have a lack of contribution docs.

I appreciate all the heavy lifting you have done for the two modules! I'm going to do the same for the `netbox_tag` PR as it just needs minor work that I can get done tonight.

Again I really appreciate your help on these and I'll leave you commits attached to make sure you get credit for these!

PR #371, opened 2020-10-31
Correct labels from 'prefix' to 'interface'

brandomando:
Some of the descriptions were copied from the `netbox_prefix` module into the `netbox_device_interface` module. This PR is just to correct those.

### DIFF
--- a/docs/plugins/modules/netbox_device_interface/netbox.netbox.netbox_device_interface_module.rst
+++ b/docs/plugins/modules/netbox_device_interface/netbox.netbox.netbox_device_interface_module.rst
@@ -81,7 +81,7 @@ Parameters
                                 <td>
                                                                                                                                                             </td>
                                                                 <td>
-                                            <div>Defines the prefix configuration</div>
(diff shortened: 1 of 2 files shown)


docs/plugins/modules/netbox_device_interface/netbox.netbox.netbox_device_interface_module.rst
```diff
@@ -81,7 +81,7 @@ Parameters
                                 <td>
                                                                                                                                                             </td>
                                                                 <td>
-                                            <div>Defines the prefix configuration</div>
+                                            <div>Defines the interface configuration</div>
```
FragmentedPacket, line 84: These are actually generated from the modules themself.

Revert these changes and update within the module itself and then I typically generate new docs on each release and these changes will be made in these `.rst` files.

brandomando: I updated the changes in the docstring in the module and reverted the changes to the docs file.

PR #380, opened 2020-11-25
Added comments option for netbox_virtual_machine

vkirovskiy:
Added comments option

### DIFF
--- a/plugins/modules/netbox_virtual_machine.py
+++ b/plugins/modules/netbox_virtual_machine.py
@@ -123,7 +123,6 @@ options:
           - Comments of the virtual machine
         required: false
         type: str
-      custom_fields:
     required: true
   state:
     description:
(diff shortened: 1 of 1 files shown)


plugins/modules/netbox_virtual_machine.py
```diff
@@ -118,6 +118,12 @@
           - configuration context of the virtual machine
         required: false
         type: dict
+      comments:
+        description:
+          - Comments of the virtual machine
+        required: false
+        type: str
+      custom_fields:
```
FragmentedPacket, line 126: Do you mind removing L126?

FragmentedPacket (changes requested):

plugins/modules/netbox_virtual_machine.py
vkirovskiy: Oh, sure, this field is appeared by wrong copy-paste
Fixed

FragmentedPacket (approved):

PR #383, opened 2020-12-04
Add virtual_chassis_name to inventory plugin

### LINKED ISSUE #382: Enhancement: Add option to use the virtual chassis name as inventory_hostname
##### ISSUE TYPE

 - Feature Idea

###### Ansible:

2.9 or 2.10

###### Netbox:

2.9.10

###### Collection:

devel

##### SUMMARY

The inventory plugin returns virtual chassis masters and uses the master's name as the host name (inventory_hostname). I'd like to use the virtual chassis name instead as the name in the inventory. Adding an option can pretty easily add that. I will be submitting a PR.

##### STEPS TO REPRODUCE

Example inventory options file:
```yaml
api_endpoint: https://netbox.example.com

virtual_chassis_name: yes
```

##### EXPECTED RESULTS

New behavior would only change the host name to a master device to the virtual chassis name. Nothing else in the host vars would change. For example the display name would still show the master device name. The tags would still be from the master device.

nahun:
Adds an option to use the virtual chassis name as the host's name (inventory_hostname). Fixes #382.


FragmentedPacket:
@nahun I have update the testing to use GitHub Actions due to Travis-CI not being free any longer.

if you want to rebase off of `devel` and the tests should show up and I can help with any issues.

FragmentedPacket (approved):

PR #391, opened 2020-12-12
Updates to look into the environment for NetBox info

### LINKED ISSUE #390: Lookup Plugin Support Environment Lookup
##### ISSUE TYPE

 - Feature Idea

###### Ansible:
2.10.4

###### Netbox:
2.9.10

###### Collection:
1.1.0

##### SUMMARY
To support the use of environment variables for the NetBox URL and NetBox Token, which will help clean up code

##### STEPS TO REPRODUCE
Set the environment
Look to get at the data without having to add the key/value pair for NetBox

```yaml
- name: "GET SITES WITH NB_QUERY"
      set_fact:
        sites: "{{ query('netbox.netbox.nb_lookup', 'sites') }}"
```

##### EXPECTED RESULTS

Would expect to get the data

##### ACTUAL RESULTS

Get an error that these were not set.

jvanderaa:
Fixes #390 

Adds the lookup of the environment variables for NetBox to the lookup plugin.


plugins/lookup/nb_lookup.py
```diff
@@ -283,8 +290,8 @@ class LookupModule(LookupBase):
 
     def run(self, terms, variables=None, **kwargs):
 
-        netbox_api_token = kwargs.get("token")
-        netbox_api_endpoint = kwargs.get("api_endpoint")
+        netbox_api_token = kwargs.get("token") or os.getenv("NETBOX_TOKEN")
+        netbox_api_endpoint = kwargs.get("api_endpoint") or os.getenv("NETBOX_API") or os.getenv("NETBOX_TOKEN")
```
FragmentedPacket, line 294: I assume the last or should be `NETBOX_URL`, but should do that since the inventory plugin only uses the `NETBOX_API` environment. Do you mind adding a second option for token as well to match the inventory plugin? `NETBOX_API_KEY`.

FragmentedPacket (changes requested):

plugins/lookup/nb_lookup.py
jvanderaa: Makes perfect sense. My bad for missing it!

FragmentedPacket (approved):

PR #412, opened 2021-01-11
updated ip doc to show family is depricated

ryanmerolle:
The RTD did not show why the family option is being depreciated, but the module code did.  This is now fixed.


FragmentedPacket:
@ryanmerolle I'm not sure this is the right solution. Those docs probably shouldn't even be committed as RTD does this behind the scenes and the docs are primarily auto-generated.

If you don't mind keeping this open though so we can see how we should do this with the way Ansible generates documentation (this is on my and not you), but want to track it somehow.

FragmentedPacket:
@ryanmerolle Hey man. Thanks for pointing this out. My last statement was incorrect as it wasn't apparently building them via RTD. I went ahead and just have them built and managed via Git since it wasn't stable for RTD.

The docs have been updated in #427 and is generated from the actual module docs instead of the `.rst` files.

Thanks for finding this!

PR #429, opened 2021-02-01
Add description field to argument_spec

pkorovin:
Existing playbooks fail with missing description field.

### DIFF
--- a/plugins/modules/netbox_device_role.py
+++ b/plugins/modules/netbox_device_role.py
@@ -142,6 +142,7 @@ def main():
                 required=True,
                 options=dict(
                     name=dict(required=True, type="str"),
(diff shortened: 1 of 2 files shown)


FragmentedPacket (changes requested):
Both of these are failing due to Ansible Sanity tests as argument specs should exist in the `DOCUMENTATION` string and vice versa.

If you add them to the `DOCUMENTATION` string, the tests should pass.

https://netbox-ansible-collection.readthedocs.io/en/latest/getting_started/contributing/modules/update_module.html#new-option-doesn-t-correlate-to-different-netbox-models

FragmentedPacket (approved):

PR #434, opened 2021-02-05
Set ansible_host to use dns_name

autogun:
In continuation to https://github.com/netbox-community/ansible_modules/pull/394,
If `primary_ip` exists on the device/vm, `ansible_host` is set to IP address.

In https://github.com/netbox-community/ansible_modules/discussions/431, was suggested to override `ansible_host` with `{{ dns_name }}` but that doesn't work (at least in my tests).

PR sets `ansible_host` to FQDN extracted from `primary_ip`  when `dns_name: True`.
This way I was successfully managed use nb_inventory module with Windows machine using Kerberos.


FragmentedPacket:
@autogun I appreciate the PR! I think we need to modify this a little bit.

I think we need to add a new option as something as: `ansible_host_dns_name` and then add if that var is true to extract the `dns_name` and then set as `ansible_host` var.

You've got most of this, but I think you'll have to make a few changes.

One being on L430 I believe to add to the `if self.interfaces or self.dns_name or self.ansible_host_dns_name` there for this new option.

And then the same thing as above, but on L1066 for the lookups.

I'll comment on the changed code though as to what changes need to be made.

plugins/inventory/nb_inventory.py
```diff
@@ -1369,6 +1369,13 @@ def _fill_host_variables(self, host, hostname):
         if extracted_primary_ip:
             self.inventory.set_variable(hostname, "ansible_host", extracted_primary_ip)
 
+        if self.dns_name:
```
FragmentedPacket, line 1372: I would structure this the same as the ones above and below it.

```suggestion
        if self.ansible_host_dns_name:
            extracted_dns_name = self.extract_dns_name(host=host)
            if extracted_dns_name:
                self.inventory.set_variable(
                    hostname, "ansible_host", extracted_dns_name
                )
```

Can you also update the tests as well? You can just update one of the test files for this new option to make sure it's setting the `ansible_host` with what is fetched as the `dns_name`.

FragmentedPacket (changes requested):

autogun:
@FragmentedPacket thanks for the feedback! Hope I'm executing this right. Maybe some more clarification is required around the tests thought.

FragmentedPacket:
@autogun I haven't messed with inventory testing too much. I may have to clone down your changes and muck with them locally first. I will keep you posted on my findings.

FragmentedPacket:
@autogun I submitted a PR into your `devel` branch with the fixes, etc.

Let me know if you have any questions, but if you merge that in, it should update this PR, and tests should pass.

autogun:
@FragmentedPacket Thanks a bunch for the assistance!

FragmentedPacket (approved):

PR #442, opened 2021-02-18
Fetch host/group vars before compose step (Resolves #439)

### LINKED ISSUE #439: Bug Report: Compose not picking up inventory_hostname field
##### ISSUE TYPE

 - Bug Report

###### Ansible:
v2.9.12

###### Netbox:
v2.9.4

###### Collection:
v2.1.0

##### SUMMARY
When using the compose functionality to add a field the `inventory_hostname` variable is not accessible.

Usually this is needed in cases that you want to add a field based on the `inventory_hostname` mutations and this is supported by the builtin constructed plugin. 

##### STEPS TO REPRODUCE
Add a compose field in the netbox configuration

```yaml
compose:
  compose_field: inventory_hostname
```

Then run ansible-inventory to see if the field is added to the host variables.

```bash
ansible-inventory -vvvv --host name-of-host
```

##### EXPECTED RESULTS

```yaml
{
    "ansible_host": "x.x.x.x",
    "compose_field": "name-of-host"
    "custom_fields": {},
    ...
}
```

##### ACTUAL RESULTS

```yaml
{
    "ansible_host": "x.x.x.x",
    "custom_fields": {},
    ...
}
```

dmavrommatis:
Added support of having `inventory_hostname` available on the compose step to have same functionality as constructed plugin.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -263,6 +263,7 @@ from ansible_collections.ansible.netcommon.plugins.module_utils.compat.ipaddress
 from ansible.utils.vars import combine_vars
 from ansible.inventory.helpers import get_group_vars
 
+
 class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
     NAME = "netbox.netbox.nb_inventory"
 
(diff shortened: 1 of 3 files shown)


FragmentedPacket:
I'll look into this, but it seems to be breaking the group_by functionality from the looks of the integration testing.

dmavrommatis:
> I'll look into this, but it seems to be breaking the group_by functionality from the looks of the integration testing.

I've run the integration tests locally on the master branch without my changes and it is getting the same error. Not really sure why on the master it replaces invalid characters while on my branch doesn't.
```
Master:
Replacing invalid character(s) "{'-'}" in group name (test-rack-role)

Checked out locally on bb98ce3:
Not replacing invalid character(s) "{'-'}" in group name (test-rack-role)
```

dmavrommatis:
closing this as it introduced other issues

PR #450, opened 2021-03-02
Enchancement: add parent_tenant_group suboption for netbox_tenant_group

pkorovin:


FragmentedPacket (changes requested):
Do you mind adding the parent option to both versions of the integration tests?

PR #479, opened 2021-03-29
Added 'label' field to interface

netixx:
Allow using the 'label' field in netbox_device_interface module


netixx:
I added the option in the docs of `netbox_interface` and in the code and docs for `netbox_device_interface`.

I guess there is work in progress to migrate from the latter to the former, with a more flexible `data` argument (avoiding precisely this type of pull request).

Please let me know if this is no longer required in the long term.

FragmentedPacket:
@netixx Thanks for your contribution!

PR #510, opened 2021-05-08
Full locations functionality in inventory plugin for 2.11

### LINKED ISSUE #492: Inventory plugin broken after upgrading Netbox to 2.11.x
##### ISSUE TYPE
 - Bug Report

###### Ansible:
2.9.16

###### Netbox:
2.11.0

###### Collection:
3.0.0

##### SUMMARY
After upgrading netbox from 2.10.6 to 2.11.0 the inventory plugin to longer works. This is caused by a rename of "rack-groups" into "locations" in Netbox version 2.11.x

##### STEPS TO REPRODUCE

- Create an inventory file with an "api_endpoint" that points to a Netbox 2.11.0 server.

- Run ansible-inventory-3 --list

```
ansible-inventory-3 --list
```

##### EXPECTED RESULTS
A list with my inventory.

##### ACTUAL RESULTS
An warning that no inventory was parsed

##### Solution
In nb_inventory.py on line 814 change:
```
"/api/dcim/rack-groups/?limit=0"
```
into:
```
"/api/dcim/locations/?limit=0"
```

And it will work again. Since the entire module has been renamed it may be elegant to replace "rack-groups" with "locations" in other places as well.

For more information check "breaking changes" in the Netbox 2.11.0 release notes:
https://github.com/netbox-community/netbox/releases

nahun:
Use locations instead of rack groups in 2.11. This still allows 2.10 or lower to function with rack groups.

Fixes #492 

- Will use locations or rack groups depending on the API version
- Adds locations to their specific site group as children
- Creates parent relationship with nested locations just like regions
- Devices are added to locations if assigned to a location. They are assigned to sites if **not** assigned to a location.
- Will throw an error if `location` is added to `group_by` and API version is 2.10 or lower
- Will throw an error if `rack_group` is added to `group_by` and API version is 2.11 or higher

I have not added any testing yet, but wanted to at least get this out there.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -1595,7 +1595,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                     group=self.location_group_names[host["location"]["id"]],
                     host=hostname,
                 )
-            else:
+            elif self.site_group_names and host.get("site"):
                 # Add host to site group when host is NOT assigned to a location
                 self.inventory.add_host(
                     group=self.site_group_names[host["site"]["id"]], host=hostname,
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -724,13 +724,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         # - it will have 1 element if the device's location has no parent
         # - it will have multiple elements if the location has a parent location
 
-        location = host.get("location", None)
-        if not isinstance(location, dict):
-            # Device has no location
-            return []
-
-        location_id = location.get("id", None)
-        if location_id is None:
+        try:
+            location_id = host["location"]["id"]
+        except (KeyError, TypeError):
             # Device has no location
             return []
 
@@ -847,7 +843,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
         # Dictionary of location id to parent location id
         self.locations_parent_lookup = dict(
-            filter(lambda x: x is not None, map(get_location_parent, locations))
+            filter(None, map(get_location_parent, locations))
         )
         # Location to site lookup
         self.locations_site_lookup = dict(map(get_location_site, locations))
@@ -1589,11 +1585,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             self.add_host_to_groups(host=host, hostname=hostname)
 
             # Special processing for sites and locations as those groups were already created
-            if (
-                getattr(self, "location_group_names", None)
-                and host.get("location")
-                and host["location"].get("id")
-            ):
+            if getattr(self, "location_group_names", None) and host.get("location"):
                 # Add host to location group when host is assigned to the location
                 self.inventory.add_host(
                     group=self.location_group_names[host["location"]["id"]],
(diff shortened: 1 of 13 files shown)


FragmentedPacket:
I'm attempting to fix integration tests for 2.11 now so hopefully i'll get that done today so you can rebase and not have too much work to get them working.

FragmentedPacket:
Is this waiting for me?

If you rebase from `devel` (I need to change that lol) the testing for 2.11 should pass with hopefully minor tweaks for your changes.

nahun:
OK thanks, I'll try to get this done today or Monday.

nahun:
@FragmentedPacket, tests have been updated and the PR should be good to go.

plugins/inventory/nb_inventory.py
```diff
@@ -704,6 +717,29 @@ def extract_regions(self, host):
...
+
+        location = host.get("location", None)
+        if not isinstance(location, dict):
+            # Device has no location
+            return []
+
+        location_id = location.get("id", None)
+        if location_id is None:
+            # Device has no location
+            return []
```
FragmentedPacket, line 735: I think this can be simplified. `location` is `None` if not set so can just return the empty list if it's `None`.

plugins/inventory/nb_inventory.py
```diff
@@ -787,6 +823,35 @@ def get_region_parent(region):
...
+            except Exception:
+                return (location["id"], None)
+
+        def get_location_site(location):
+            # Locations MUST be assigned to a site
+            return (location["id"], location["site"]["id"])
+
+        # Dictionary of location id to parent location id
+        self.locations_parent_lookup = dict(
+            filter(lambda x: x is not None, map(get_location_parent, locations))
```
FragmentedPacket, line 850:
```suggestion
            filter(None, map(get_location_parent, locations))
```

plugins/inventory/nb_inventory.py
```diff
@@ -1488,9 +1588,22 @@ def main(self):
...
             self.add_host_to_groups(host=host, hostname=hostname)
 
-        # Create groups for regions, containing the site groups
-        if "region" in self.group_by:
-            self._add_region_groups()
+            # Special processing for sites and locations as those groups were already created
+            if (
+                getattr(self, "location_group_names", None)
+                and host.get("location")
+                and host["location"].get("id")
```
FragmentedPacket, line 1595: This may apply to the same comments above, should be able to get rid of L1595.

FragmentedPacket (changes requested):
Thanks a lot for taking the time to update this! Great work and just a few nitpicks and then we should be good to merge it!

plugins/inventory/nb_inventory.py
nahun: I just copied the `extract_regions()` function, so I guess I didn't look at it to closely. Are you saying if `host['location']` is not `None` it will always have an `'id'` key set so we don't need to check it?

I can change to something like this to cover both still:
```python
location_id = host.get("location", {}).get("id", None)
if not location_id:
    # Device has no location
    return []
```
Or are you thinking just this and not bother with `'id'`:
```python
if not host.get("location", None):
    # Device has no location
    return []
location_id = host["location"]["id"]
```

FragmentedPacket: Correct. I think that should be the case.

nahun: I went with
```python
try:
    location_id = host["location"]["id"]
except (KeyError, TypeError):
    # Device has no location
    return []
```
which is probably more pythonic...

bluikko:
I do not see `location` in the `compose` functionality.

PR #537, opened 2021-06-09
Fix plugins/inventory -> options/group_by

felix-russell:
The documentation for a plugin's `options` are located [here]( https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec).

Changing this to `elements` allows users to select "Any of" the options in the list rather than "The whole list (or not)" As it stands the `default: []` prevents the plugin from loading at all.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -114,7 +114,7 @@ DOCUMENTATION = """
                 - I(rack_group) is supported on NetBox versions 2.10 or lower only
                 - I(location) is supported on NetBox versions 2.11 or higher only
             type: list
-            choices:
+            elements:
                 - sites
                 - site
                 - location


2bithacker (approved):

ryanmerolle:
@felix-russell Can you better explain this?  I never had an issue loading the plugin, you give no error message or link to an issue reproducing the issue.  I am trying to cut a release, and I'm having to go down the rabbit hole of this PR because it breaks the antsibull docs auto build as it is stating:

```
        1 validation error for PluginDocSchema
        doc -> options -> group_by -> elements
          str type expected (type=type_error.str)
```

Nautobot ansible foked this repo, and they kept choices also.

ryanmerolle:
For now I am reverting this change given it seems to contradict the ansible docs & the antsibull collection process does not recognize it.

PR #546, opened 2021-06-16
Bump pydantic 1.7.4 (CVE-2021-29510), urllib3 1.26.5 (CVE-2021-33503)

aminvakil:


aminvakil:
Oops, that file is generated automatically.

PR #558, opened 2021-07-14
Quick fix to support ansible-core 2.11

### LINKED ISSUE #529: multiple modules, same error:  "msg": "state is present but all of the following are missing: name"
##### ISSUE TYPE
 - Bug Report

###### Ansible:
ansible [core 2.11.0]

###### Netbox:
2.10.4

###### Collection:
3.1.1,  But also tested with:
3.1.0
3.0.0

(all with same result)

##### SUMMARY
I get the same error, regardless of module I attempt to use: (Tested with netbox_device, netbox_site, netbox_manufacturer)

`state is present but all of the following are missing: name`

##### STEPS TO REPRODUCE
Example from official documentation:
```yaml
cat roles/netbox/tasks/main.yml 
# tasks file for netblox
---

- name: "Example for state: present"
  netbox.netbox.netbox_device:
    netbox_url: "{{ lookup('env', 'NETBOX_URL') }}"
    netbox_token: "{{ lookup('env', 'NETBOX_API_KEY') }}"
    data:
      name: "Test Device"
      device_type: "C9410R"
      device_role: "Core Switch"
      site: "Main"
    state: present
```

Being called from a playbook with:
```yaml
---

- name: Add items to netbox
  hosts: localhost
  connection: local
  roles:
    - netbox
```

##### EXPECTED RESULTS
An example cut/pasted from the docs (obviously replacing the URL and API key) should result in a device called "Test Device" being created.

##### ACTUAL RESULTS
I get an error 
```
fatal: [localhost]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "data": {
                "asset_tag": null,
                "cluster": null,
                "comments": null,
                "custom_fields": null,
                "device_role": "Core Switch",
                "device_type": "C9410R",
                "face": null,
                "local_context_data": null,
                "name": "Test Device",
                "platform": null,
                "position": null,
                "primary_ip4": null,
                "primary_ip6": null,
                "rack": null,
                "serial": null,
                "site": "Main",
                "status": null,
                "tags": null,
                "tenant": null,
                "vc_position": null,
                "vc_priority": null,
                "virtual_chassis": null
            },
            "netbox_token": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "netbox_url": "https://<<edited>>,
            "query_params": null,
            "state": "present",
            "validate_certs": true
        }
    },
    "msg": "state is present but all of the following are missing: name"
}

```
As you can see from the task definition, and resulting debug output, "name" is clearly defined and parsed.
However, I still get the error message.

It's worth noting, that I took all of the "data" elements from this debug output, an dmade sure to populate *all* fields (assuming the error was about some other field other than "name", but it resulted in the exact same response.
Additionally worth noting, I'm only showing an example netbox_device, but the same error is happening with multiple as mentioned above.

As an example of one where I populated all fields as a test (in this case, netbox_site)
```
fatal: [localhost]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "data": {
                "asn": "Foo",
                "comments": "Foo",
                "contact_email": "Foo",
                "contact_name": "Foo",
                "contact_phone": "Foo",
                "custom_fields": "Foo",
                "description": "Foo",
                "facility": "Foo",
                "latitude": "Foo",
                "longitude": "Foo",
                "name": "Foo",
                "physical_address": "Foo",
                "region": "Foo",
                "shipping_address": "Foo",
                "slug": "Foo",
                "status": "Foo",
                "tags": [
                    "Foo"
                ],
                "tenant": "Foo",
                "time_zone": "Foo"
            },
            "netbox_token": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "netbox_url": "https://<<edited>>",
            "query_params": null,
            "state": "present",
            "validate_certs": true
        }
    },
    "msg": "state is present but all of the following are missing: name"
}
```

I DO see a note in the documentation `These same fields are required when creating a device via the netbox_device module, but with the important addition of name.` However, as you can see.. it's being provided.

nahun:
Fixes #529

This is a quick fix so that the modules will work with ansible-core 2.11.

The validation process has changed as detailed in #529. This change runs the validation checks manually.

The params have to be loaded manually as the AnsibleModule adds the default values by the time we can run the manual checks.

This should not be a long term fix.


felixfontein:
Looks good to me. What's still needed though:
1. You need to add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) to https://github.com/netbox-community/ansible_modules/tree/devel/changelogs/fragments
2. A lot of ignore.txt entries need to be removed from https://github.com/netbox-community/ansible_modules/blob/devel/tests/sanity/ignore-2.10.txt, and it looks like others should be added.

nahun:
@felixfontein Thanks, I've updated the PR.

The sanity tests are still failing as I don't think its in this PR's scope to fix the `cert` argspec errors. This was added in #534 and not done completely. It was missing a lot with the argspec docs and I believe it plain just breaks all modules as well if you don't specify `cert` option when calling any module in a playbook.

nahun:
Oh actually, the `cert` errors were already fixed in #573 by Ryan.

felixfontein:
> @felixfontein Thanks, I've updated the PR.
> 
> The sanity tests are still failing as I don't think its in this PR's scope to fix the `cert` argspec errors. This was added in #534 and not done completely. It was missing a lot with the argspec docs and I believe it plain just breaks all modules as well if you don't specify `cert` option when calling any module in a playbook.

I fully agree with this. (That PR also misses some more things, like marking the new options with `version_added: ...`.)

felixfontein (approved):

gundalow (approved):
Although this doesn't totally fix CI, it does improve the situation. 
I think it would be good to merge this as-is, then we can continue wth #573 (or a different PR)

PR #591, opened 2021-09-15
Fixes #583: Move to Ansible-core for CI tests

### LINKED ISSUE #583: CI does not run sanity tests with ansible-core 2.11 and ansible-core's devel branch
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
It currently only runs these tests with ansible-base 2.10, which is not a good idea at all.

Check out the testing requirements for inclusion in Ansible: https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#ci-testing

rodvand:
- Moved from ansible-base to ansible-core
- Removed form_factor from interface and family from ip_address as they were marked deprecated/removed
- Add test for pynetbox/requests library in nb_lookup.py
- Substituted list comprehension to match sanity test
- Removed family from test data to match removal from module
- Ignore pylint test on nb_lookup due to bug https://github.com/PyCQA/pylint/issues/3507 in pylint 2.6.0 which is used in ansible-test

### DIFF
--- a/plugins/modules/netbox_device_interface.py
+++ b/plugins/modules/netbox_device_interface.py
@@ -63,14 +63,6 @@ options:
           - Physical label of the interface
         required: false
         type: str
-      form_factor:
-        description:
-          - |
-            Form factor of the interface:
-            ex. 1000Base-T (1GE), Virtual, 10GBASE-T (10GE)
-            This has to be specified exactly as what is found within UI
-        required: false
-        type: raw
       type:
         description:
           - |
@@ -298,12 +290,6 @@ def main():
                 options=dict(
                     device=dict(required=False, type="raw"),
                     name=dict(required=True, type="str"),
-                    form_factor=dict(
-                        required=False, 
-                        type="raw", 
-                        removed_in_version="0.3.0", 
-                        removed_in_collection="netbox.netbox"
-                    ),
                     label=dict(required=False, type="str"),
                     type=dict(required=False, type="str"),
                     enabled=dict(required=False, type="bool"),
--- a/plugins/modules/netbox_device_interface.py
+++ b/plugins/modules/netbox_device_interface.py
@@ -299,7 +299,10 @@ def main():
                     device=dict(required=False, type="raw"),
                     name=dict(required=True, type="str"),
                     form_factor=dict(
-                        required=False, type="raw", removed_in_version="0.3.0"
+                        required=False, 
+                        type="raw", 
+                        removed_in_version="0.3.0", 
+                        removed_in_collection="netbox.netbox"
                     ),
                     label=dict(required=False, type="str"),
                     type=dict(required=False, type="str"),
--- a/plugins/modules/netbox_device_interface.py
+++ b/plugins/modules/netbox_device_interface.py
@@ -63,6 +63,14 @@ options:
           - Physical label of the interface
         required: false
         type: str
+      form_factor:
+         description:
+           - |
+             Form factor of the interface:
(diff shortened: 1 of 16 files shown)


ryanmerolle:
Closes #583

ryanmerolle (approved):

sc68cal (dismissed):

plugins/modules/netbox_device_interface.py
```diff
@@ -298,9 +290,6 @@ def main():
                 options=dict(
                     device=dict(required=False, type="raw"),
                     name=dict(required=True, type="str"),
-                    form_factor=dict(
```
felixfontein, line 301: Removing options requires a new major release, which I would really avoid for now until you have a new minor (or patch) release that fixes the current issues. A new major release has to wait for Ansible 5 and cannot be included in the next Ansible 4 release (which will happen ~next Tuesday BTW).

plugins/modules/netbox_ip_address.py
```diff
@@ -323,12 +312,6 @@ def main():
                 type="dict",
                 required=True,
                 options=dict(
-                    family=dict(
```
felixfontein, line 326: Same here.

felixfontein (commented):
I'd suggest to remove the removals from this PR and wait with them for the next major release.

felixfontein:
(Also you'd need a `removed_features` changelog fragment.)

plugins/modules/netbox_device_interface.py
rodvand: Thanks @felixfontein , this is great feedback! I did try to add the `'removed_from_collection': 'netbox.netbox'` but then the tests said `ERROR: plugins/modules/netbox_device_interface.py:0:0: collection-deprecated-version: Argument 'form_factor' in argument_spec found in data has a deprecated removed_in_version '0.3.0', i.e. the version is less than or equal to the current version of this collection (3.1.1)`.

Does this mean it should really have been removed before? And should I now just bump the `removed_in_version` to 4.0.0 to resolve this as smooth as possible? And then I guess the tests will prompt me if I'm not removing it from version 4.0.0?

felixfontein: Yes, it should have been removed in 1.0.0. (Or maybe the value '0.3.0' is also wrong, maybe it was the version when the deprecation was added?)

In any case, you could either add an ignore.txt entry for this, or replace 0.3.0 by 4.0.0, as you suggested. In the latter case, you (or whoever else) will be reminded once you change the version to 4.0.0 and run sanity checks.

ryanmerolle: This is super useful feedback!  Thanks @felixfontein !

plugins/modules/netbox_ip_address.py
ryanmerolle: Again Thanks!

ryanmerolle:
@rodvand, do you think you would be able to address @felixfontein comments.  I will work with @FragmentedPacket and @jeremystretch to try to get a release cut and pushed to ansible galaxy before the next release of Ansible 4 this up coming Tuesday.

rodvand:
> @rodvand, do you think you would be able to address @felixfontein comments. I will work with @FragmentedPacket and @jeremystretch to try to get a release cut and pushed to ansible galaxy before the next release of Ansible 4 this up coming Tuesday.

Working on it.

ryanmerolle (approved):
Looks good to me.

@felixfontein @gundalow @rodvand any final comments or items before I merge?

felixfontein (approved):
LGTM (as far as I can judge :) )

PR #600, opened 2021-09-23
 Using Primary IP as inventory_hostname

aventrax:
Adding an option to use primary IP as inventory_hostname in place of the netbox device name.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -1583,7 +1583,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 # Device is part of a virtual chassis, but is not the master
                 continue
 
-            if self.ansible_inventory_primary_ip:
+            if self.ansible_inventory_primary_ip and self.extract_primary_ip(host=host):
                 hostname = self.extract_primary_ip(host=host)
             else:
                 hostname = self.extract_name(host=host)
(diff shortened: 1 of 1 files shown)


ryanmerolle:
@aventrax the CI failed on some simple black tests.  Feel free to peruse them and update accordingly

aventrax:
@ryanmerolle I run black but it modifies even something not touched by me. What should I do? (row 451, it reformat "location", "rack_group", "services", "interfaces" and "dns_name").

aventrax (commented):
I manually changed what seems to be the issue with indentation. Hope it works.

plugins/inventory/nb_inventory.py
```diff
@@ -1578,7 +1583,11 @@ def main(self):
                 # Device is part of a virtual chassis, but is not the master
                 continue
 
-            hostname = self.extract_name(host=host)
+            if self.ansible_inventory_primary_ip and self.extract_primary_ip(host=host):
+                hostname = self.extract_primary_ip(host=host)
```
sc68cal, line 1587: This is a little hacky. I don't know if I like setting a variable named `hostname` with a raw IP address in some cases.

sc68cal:
I think I would prefer that this set the [`ansible_host`] setting for a host.

plugins/inventory/nb_inventory.py
aventrax: I agree with you. Unfortunately Ansible itself has IP addresses straight on the inventory in their documentation, so any var used for that could contain an hostname or an IP.

sc68cal: Yes but my nit is that in most of the Ansible internals, they use `host` - which could either be an IP or a DNS name. You are assigning an IP address to a variable called `hostname` which, in my opinion, breaks the meaning of the variable.

Issue #507, opened 2021-05-06
Add parent_interface to netbox_device_interface


andybro19:
##### ISSUE TYPE

 - Feature Idea

###### Netbox:

2.11

##### SUMMARY

Since netbox 2.11 you can assign interfaces to a parent interface and parent_interface should be added as an option for netbox_device_interface. E.g:

```yaml
- name: Assign interface to parent
  netbox.netbox.netbox_device_interface:
    data:
      device: device
      name: interface
      type: Virtual
      parent_interface: ParentInterface
    state: present
```

PR #604, opened 2021-09-27
This allows the addition of the parent to an interface.

### LINKED ISSUE #549: Add parent_vm_interface to netbox_vm_interface
##### ISSUE TYPE

 - Feature Idea

###### Netbox:

2.11

##### SUMMARY

Since netbox 2.11 you can assign vm interfaces to a parent interface and parent_vm_interface should be added as an option for netbox_vm_interface. E.g:

```yaml
- name: Assign interface to parent
  netbox.netbox.netbox_vm_interface:
    data:
      virtual_machine: vm
      name: interface
      parent_vm_interface:
        virtual_machine: vm
        name: ParentInterface
    state: present
```

ThomasADavis:
This fixes #549 and #507 

This allows the parent of a interface to be assigned.

Both virtual (vm) and real (device)

for real devices, the label is 'parent_interface'.

for vm devices, the label is 'parent_vm_interface'.

This is done this way for several reasons.

1) prevention of a collision with the 'parent' label in other modules.
2) assignment of the correct netbox API.

The parent can only be on the same device/VM. Trying to assign a different interface from a different device will fail.  The child interface also must be a virtual interface on real devices, they already are virtual on VM's.

This has been tested in my environment for several months now.

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -177,6 +177,7 @@ CONVERT_TO_ID = {
     "nat_inside": "ip_addresses",
     "nat_outside": "ip_addresses",
     "platform": "platforms",
+    "parent": "interfaces",
     "parent_region": "regions",
     "parent_tenant_group": "tenant_groups",
     "power_panel": "power_panels",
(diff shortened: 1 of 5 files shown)


rodvand:
Hi @ThomasADavis! Thanks for adding the PR! Could you investigate as to why the CI tests are failing? And maybe look into adding an integration test this addition?

ThomasADavis:
yea.. netbox_prefix uses the same keyword...  so it bombs on that.

ThomasADavis:
as a note - this is working now.

ryanmerolle (approved):

PR #617, opened 2021-10-21
Cache openapi dump locally

kr3ator:
Hi, I'd like to propose a change to the _nb_inventory_ that speeds up the fetching of data on subsequent runs.

Currently, the NetBox API docs are always downloaded at each inventory script run.
This takes a lot of time to finish as the file is now over 60000 lines long.
Enabling the cache option is not a good solution for this as it will also cache all the objects gathered by the script.

My solution leverages a new _status_ API endpoint added in NetBox 2.10 which returns NetBox version among other parameters.
The API version from the GET status call is then compared with the locally dumped JSON with API docs. If the versions match then locally cached JSON file is loaded. In any other scenario the API docs are fetched as usual.

When implemented my test inventory was gathered in 3s whereas with original plugin it took 22s.

Happy to discuss this further.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -258,6 +258,7 @@ keyed_groups:
 import json
 import uuid
 import math
+import os
 from copy import deepcopy
 from functools import partial
 from sys import version as python_version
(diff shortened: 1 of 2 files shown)


kr3ator:
@sc68cal I've just pushed the tests.
I wanted to have just a single decorator with the ```mock_open(read_data="""{"info": {"version": "2.0"}}""")``` instead of double 'with' sandwitch inside the test function. Normally it should work as ```json.load()``` uses the ```.read()``` method on the file handler but  the ```json.load()``` complained that ```TypeError: the JSON object must be str, bytes or bytearray, not MagicMock```

I'm open for suggestions.

sc68cal:
yeah, that's the fun of `Mock`, but honestly I'm just pleased to see a test for it, not going to dwell too much on the work it takes to make `Mock` happy. So great job, I'm 👍 on this PR

sc68cal (approved):

PR #631, opened 2021-10-31
Fixes #618 Ensure NetBox site module is idempotent between NetBox versions

### LINKED ISSUE #618: netbox_site: Not idempotent on v3.0 when using latitude/longitude.
##### ISSUE TYPE

 - Bug Report

###### Ansible:

v2.11.6

###### Netbox:

v3.0.8

###### Collection:

v3.2.0

##### SUMMARY

NetBox v3.0 uses a float instead of a string to store the latitude and longitude. The module uses a string and is no longer idempotent with v3.0.

##### STEPS TO REPRODUCE

```yaml
---
- name: "Test Netbox site module"
  connection: local
  hosts: localhost
  gather_facts: False
  tasks:
    - name: Create a site
      netbox_site:
        data:
          name: Test Site
          latitude: 22.123
        state: present
```

##### EXPECTED RESULTS

The site is not changed.

##### ACTUAL RESULTS

The module reports changed.

```
TASK [Create a site] ******************************************************************************************
--- before
+++ after
@@ -1,3 +1,3 @@
 {
-    "latitude": 22.123
+    "latitude": "22.123"
 }

changed: [localhost]
```

rodvand:
- Change latitude and longitude parameters from str to float to reflect the change in NetBox 3.0
- On versions lesser than NetBox 3.0, cast latitude and longitude to str

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -1144,6 +1144,20 @@ class NetboxModule(object):
             serialized_nb_obj["tags"] = set(serialized_nb_obj["tags"])
             updated_obj["tags"] = set(data["tags"])
 
+        # Ensure idempotency between version 2.11 and 3.0
+        if (
+            serialized_nb_obj.get("latitude")
+            and data.get("latitude")
+            and self._version_check_greater("3.0", self.version)
+        ):
+            updated_obj["latitude"] = str(data["latitude"])
+        if (
+            serialized_nb_obj.get("longitude")
+            and data.get("longitude")
+            and self._version_check_greater("3.0", self.version)
+        ):
+            updated_obj["longitude"] = str(data["longitude"])
+
         if serialized_nb_obj == updated_obj:
             return serialized_nb_obj, None
         else:
@@ -1311,7 +1325,7 @@ class NetboxAnsibleModule(AnsibleModule):
         return results
 
     def _check_required_if(self, spec, param=None):
-        """ ensure that parameters which conditionally required are present """
+        """ensure that parameters which conditionally required are present"""
         if spec is None:
             return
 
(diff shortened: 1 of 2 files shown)


plugins/module_utils/netbox_utils.py
```diff
@@ -1311,7 +1325,7 @@ def check_mutually_exclusive(self, terms, module_parameters):
         return results
 
     def _check_required_if(self, spec, param=None):
-        """ ensure that parameters which conditionally required are present """
+        """ensure that parameters which conditionally required are present"""
```
sc68cal, line 1328: this change appears to be unrelated to the subject of the PR.

plugins/module_utils/netbox_utils.py
```diff
@@ -1144,6 +1144,20 @@ def _update_netbox_object(self, data):
...
+        if (
+            serialized_nb_obj.get("latitude")
+            and data.get("latitude")
+            and self._version_check_greater("3.0", self.version)
+        ):
+            updated_obj["latitude"] = str(data["latitude"])
+        if (
+            serialized_nb_obj.get("longitude")
+            and data.get("longitude")
+            and self._version_check_greater("3.0", self.version)
```
sc68cal, line 1157: Instead of calling the `_version_check_greater()` function twice, assign the result of the function call to a local variable and use it for comparison

sc68cal (commented):
Some small nits

plugins/module_utils/netbox_utils.py
rodvand: Yeah, not sure why black suddenly decided to reformat that doc string.

plugins/module_utils/netbox_utils.py
rodvand: That's a valid point. If you agree that these sort of issues can be dealt with this way I'll do something similar for the virtual machine and its vcpu argument which changed from str to float as well.

sc68cal (approved):

PR #643, opened 2021-11-07
Prep for release 3.4.0

rodvand:
- Add version_added to new parameters
- Generate docs
- Bump galaxy version


sc68cal:
This branch appears to revert #629 , is that intended?

rodvand:
@sc68cal It's the generation of documentation which does this... ansible-doc. It also totally reverts the validate_certs change in nb_lookup wherein the documentation states True/False is to be used, and Ansible-doc generates yes/no to be the options. I guess that's why it's been wrong all this time.

sc68cal:
OK. I should have looked at things more closely before merging #629, I had this feeling it was automatically generated but didn't do any digging. Maybe it's time to look for how it gets generated. I'll try and take a look at that this week

Issue #640, opened 2021-11-04
Pull complete site data and prefixes (optionally) in Inventory module


Eric-Sim:
##### ISSUE TYPE

 - Feature Idea

###### Ansible:

All

###### NetBox:

2.11

###### Collection:

All

##### SUMMARY

Site data is not pulled by the inventory module, only the slug.
Additionally, Prefix data is not pulled. Prefixes that are assigned to a site in Netbox should be (optionally) inserted into the site data.

##### STEPS TO REPRODUCE

By default, the inventory module should include site data in the inventory within the "Site" key of devices.
Optionally, by providing "prefixes: True" to the inventory config, Prefixes along with accompanying data should be pulled and inserted into inventory under a "prefixes" key within the "sites" key.

Workflow for pulling the prefixes will follow the precedent established by the "interfaces" and "services" options.
Sites that have prefixes assigned to them will be tracked  and passed to get_resource_list_chunked() for optimization when the "fetch_all" option is False.

```
ansible-inventory -i netbox_inventory.yml --list
```

##### ACTUAL RESULTS

```

```

#### NOTES
I've already written the code to accomplish this. I'm opening this issue to be referenced by my forthcoming PR.
I just need to add documentation and update tests.

PR #646, opened 2021-11-09
Inventory module site and prefix data extended

Eric-Sim:
This PR satisfies **(Issue #640)** without causing any changes to default functionality.

Both the extension of Site Data and Prefix Data are enabled by additional parameters passed to the module.
`site_data: (boolean)`
`prefixes: (boolean)`
These changes follow the same methodology as the interfaces option.

No changes in the default behavior of the module in order to guarantee legacy applications will continue working as intended.

The site_data parameter will cause the entire data structure returned by the /api/dcim/sites API call to be captured and inserted into hostvars appropriately for each device in inventory, instead of just the slug that is collected currently.

The prefixes option extends this section of the hostvars further by nesting the datastructure of the prefix assigned to the site as well. This is done by a calling an additional endpoint /api/ipam/prefixes. 
Current optimizations are maintained with the "fetch_all" option.
Only ACTIVE prefixes, that are assigned to a site are pulled; and only for the sites that have at least one queried device assigned to them.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -847,7 +847,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 # Used by refresh_prefixes()
 
         def get_region_for_site(site):
-            # Will fail if site does not have a region defined in Netbox
+            # Will fail if site does not have a region defined in NetBox
             try:
                 return (site["id"], site["region"]["id"])
             except Exception:
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -456,26 +456,36 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         # Locations were added in 2.11 replacing rack-groups.
         if self.api_version >= version.parse("2.11"):
             extractors.update(
-                {"location": self.extract_location,}
+                {
+                    "location": self.extract_location,
+                }
             )
         else:
             extractors.update(
-                {"rack_group": self.extract_rack_group,}
+                {
+                    "rack_group": self.extract_rack_group,
+                }
             )
 
         if self.services:
             extractors.update(
-                {"services": self.extract_services,}
+                {
+                    "services": self.extract_services,
+                }
             )
 
         if self.interfaces:
             extractors.update(
-                {"interfaces": self.extract_interfaces,}
+                {
+                    "interfaces": self.extract_interfaces,
+                }
             )
 
         if self.interfaces or self.dns_name or self.ansible_host_dns_name:
             extractors.update(
-                {"dns_name": self.extract_dns_name,}
+                {
+                    "dns_name": self.extract_dns_name,
+                }
             )
 
         return extractors
@@ -602,7 +612,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
     def extract_site(self, host):
         try:
             site = self.sites_lookup[host["site"]["id"]]
-            if self.prefixes:  # If prefixes have been pulled, attach prefix to its assigned site
+            if (
+                self.prefixes
+            ):  # If prefixes have been pulled, attach prefix to its assigned site
                 prefix_id = self.prefixes_sites_lookup[site["id"]]
                 prefix = self.prefixes_lookup[prefix_id]
                 site["prefix"] = prefix
@@ -859,7 +871,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
     # Note: depends on the result of refresh_sites_lookup for self.sites_with_prefixes
     def refresh_prefixes(self):
         # Pull prefixes with "active" status only
-        url = (self.api_endpoint + "/api/ipam/prefixes?status=active")
+        url = self.api_endpoint + "/api/ipam/prefixes?status=active"
 
         if self.fetch_all:
             prefixes = self.get_resource_list(url)
@@ -1487,7 +1499,12 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         # Map site id to transformed group names
         self.site_group_names = dict()
 
-        for site_id, site_name in self.sites_lookup_slug.items():  # "Slug" only. Data not used even if pulled
+        for (
+            site_id,
+            site_name,
+        ) in (
+            self.sites_lookup_slug.items()
+        ):  # "Slug" only. Data not used even if pulled
             site_group_name = self.generate_group_name(
                 self._pluralize_group_by("site"), site_name
             )
@@ -1697,7 +1714,8 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             elif getattr(self, "site_group_names", None) and host.get("site"):
                 # Add host to site group when host is NOT assigned to a location
                 self.inventory.add_host(
-                    group=self.site_group_names[host["site"]["id"]], host=hostname,
+                    group=self.site_group_names[host["site"]["id"]],
+                    host=hostname,
                 )
 
     def parse(self, inventory, loader, path, cache=True):
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -1489,9 +1489,13 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         self.site_group_names = dict()
 
         for (site_id, site_name) in (self.sites_lookup_slug.items()):  # "Slug" only. Data not used for grouping
-            site_group_name = self.generate_group_name(self._pluralize_group_by("site"), site_name)
+            site_group_name = self.generate_group_name(
+                self._pluralize_group_by("site"), site_name
+            )
             # Add the site group to get its transformed name
-            site_transformed_group_name = self.inventory.add_group(group=site_group_name)
+            site_transformed_group_name = self.inventory.add_group(
+                group=site_group_name
+            )
             self.site_group_names[site_id] = site_transformed_group_name
 
     def _add_region_groups(self):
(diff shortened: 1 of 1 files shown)


Eric-Sim:
Please link to issue #640

plugins/inventory/nb_inventory.py
```diff
@@ -445,26 +456,36 @@ def group_extractors(self):
         # Locations were added in 2.11 replacing rack-groups.
         if self.api_version >= version.parse("2.11"):
             extractors.update(
-                {"location": self.extract_location,}
+                {
+                    "location": self.extract_location,
+                }
```
sc68cal, line 461: This change appears to be unrelated to your patch series, and should not have caused an issue with `black` - please remove this change and the other changes like this that are unrelated.

plugins/inventory/nb_inventory.py
```diff
@@ -445,26 +456,36 @@ def group_extractors(self):
...
             )
         else:
             extractors.update(
-                {"rack_group": self.extract_rack_group,}
+                {
+                    "rack_group": self.extract_rack_group,
+                }
```
sc68cal, line 467: Same issue as L461

plugins/inventory/nb_inventory.py
```diff
@@ -445,26 +456,36 @@ def group_extractors(self):
...
             )
 
         if self.services:
             extractors.update(
-                {"services": self.extract_services,}
+                {
+                    "services": self.extract_services,
+                }
```
sc68cal, line 474: Same issue as L461

plugins/inventory/nb_inventory.py
```diff
@@ -445,26 +456,36 @@ def group_extractors(self):
...
             )
 
         if self.interfaces:
             extractors.update(
-                {"interfaces": self.extract_interfaces,}
+                {
+                    "interfaces": self.extract_interfaces,
+                }
```
sc68cal, line 481: Same issue as L461

plugins/inventory/nb_inventory.py
```diff
@@ -445,26 +456,36 @@ def group_extractors(self):
...
             )
 
         if self.interfaces or self.dns_name or self.ansible_host_dns_name:
             extractors.update(
-                {"dns_name": self.extract_dns_name,}
+                {
+                    "dns_name": self.extract_dns_name,
+                }
```
sc68cal, line 488: Same issue as L461

plugins/inventory/nb_inventory.py
```diff
@@ -820,6 +868,31 @@ def get_region_for_site(site):
         # Dictionary of site id to region id
         self.sites_region_lookup = dict(map(get_region_for_site, sites))
 
+    # Note: depends on the result of refresh_sites_lookup for self.sites_with_prefixes
+    def refresh_prefixes(self):
+        # Pull prefixes with "active" status only
+        url = self.api_endpoint + "/api/ipam/prefixes?status=active"
```
sc68cal, line 874: I'm not sure about only pulling prefixes that are in `active` state. I will have to check but I don't know how often we filter things based on `status` - I am guessing the behavior is to pull everything regardless of `status`

I think as a user I'd probably want all prefixes associated and then determine on my own what do to with ones with different statuses, rather than having this bit of code decide for me.

sc68cal (commented):
Some thoughts

plugins/inventory/nb_inventory.py
Eric-Sim: I didn't notice that black had modified anything outside of the lines I changed / added.
I will get this changed back to its original formatting.

plugins/inventory/nb_inventory.py
Eric-Sim: Same fix.

plugins/inventory/nb_inventory.py
Eric-Sim: Same fix.

plugins/inventory/nb_inventory.py
Eric-Sim: Same fix.

plugins/inventory/nb_inventory.py
Eric-Sim: Same fix.

plugins/inventory/nb_inventory.py
```diff
@@ -820,6 +856,32 @@ def get_region_for_site(site):
...
+        else:
+            prefixes = self.get_resource_list_chunked(
+                api_url=url,
+                query_key="site",
+                query_values=list(self.sites_with_prefixes),
+            )
+        self.prefixes_sites_lookup = defaultdict(dict)
+        self.prefixes_lookup = defaultdict(dict)
+
+        # We are only concerned with Prefixes that have actually been assigned to sites
```
sc68cal, line 875: Is it possible to query the API and only get back prefixes assigned to this site, instead of looping through all the prefixes that are in the DB? My concern is that this will have bad performance on NetBox instances with lots of prefixes, similar to the bad performance I have experienced with `cables`

Eric-Sim: It does not look like the API provides a query for "is_assigned" or similar.

My goal here was to follow the precedent that was already set by existing code. 
The optimization of grabbing only prefixes that are relevant (by being assigned to a site, is already handled by the `get_resource_list_chunked()` function.
As you said though, it is only used if the user is not using fetch_all option.
Depending on the number of sites and prefixes, assigned and unassigned in the user's database, fetch_all may or may not be faster.

plugins/inventory/nb_inventory.py
```diff
@@ -820,6 +856,32 @@ def get_region_for_site(site):
...
+        for prefix in prefixes:
+            if prefix.get("site"):
+                prefix_id = prefix["id"]
+                site_id = prefix["site"]["id"]
+                self.prefixes_lookup[prefix_id] = prefix
+                self.prefixes_sites_lookup[site_id] = prefix_id
+            # Remove "site" attribute, as it's redundant when prefixes are assigned to site
```
sc68cal, line 882: This may still be useful to trace back to a site, if the prefix ends up getting shared with another function or something else. I don't think it's wise to delete expected attributes

Eric-Sim: This was done to follow the precedent set by the refresh_ipaddresses() function.
 In refresh_ipaddresses() the "interface" attribute is removed because this information is redundant once the whole IP Address dictionary is nested within self.ipaddresses_intf_lookup[interface_id][ip_id] and self.ipaddresses_lookup[ip_id].

I am fine with leaving this attribute, but removing it provides the user with a cleaner inventory. (less redundant info)

sc68cal: Ok, if it folllows an existing convention then that's fine

sc68cal:
approved a run through CI

Eric-Sim:
Looks like 2 of the tasks are issues with the CI pipeline itself, but the other 2 are black complaining.
I allowed this formatting in 9ff30f933053353a8b7589af7114853013538f63, but ended up undoing it when reviewing with sc68cal.

Please let me know if I need to allow black to reformat.
(the changes it wants to make do not help readability)

sc68cal:
I think we need to tweak our `black` settings, because I too do not care for that formatting that `black` is insisting on

sc68cal:
@rodvand Can you take a look at this? The problem is that `black` was complaining about a line that was not originally part of his patchset, and in the process of doing this PR he changed [this line](https://github.com/netbox-community/ansible_modules/commit/9ff30f933053353a8b7589af7114853013538f63#diff-50d60d3dd504c0a9a89108460184658acfa2c4b980f80edc78a43f68b3981e83L459) but frankly I don't think it was a related change.

Does he need to rebase?

rodvand:
Not quite sure why this happens. @Eric-Sim Do you run black locally to format the code? @sc68cal The latest black requires python 3.6.2 so we need to up our requirements before we upgrade our black version. When that's done it shouldn't be a big deal upgrading. Did a test run with a newer black and it only changed a few of the files, so not a massive re-format.

Eric-Sim:
I had run black locally to format the code. And it did apply the same formatting that the project CI/CD pipeline is trying to apply.  The problem was what  @sc68cal had pointed out, it ends up changing things that are outside the scope of what I revised; so I reverted those.

I will rebase from devel tonight.

Eric-Sim:
My branch is up to date. No new commits in Devel that I can rebase into my branch.

I have re-run black locally and committed its formatting changes.

sc68cal:
OK, something must be going on with your local `black` because it is failing on those lines in CI.

Eric-Sim:
Sorry for the slow update. 
I went ahead and manually matched the changes the the project's black wanted, and pushed.

sc68cal (approved):
This looks good to me, probably this needs to be a squash & merge operation

bluikko:
What about the functionality for `virtualization` app? This only does it for `dcim`?

For users who have both physical and virtual infrastructure a feature parity between the 2 apps would really be needed - or everything must be done for the lowest common denominator (i.e. if some feature is only for `dcim` but not `virtualization` then it could not be used).

Eric-Sim:
> What about the functionality for `virtualization` app? This only does it for `dcim`?
> 
> For users who have both physical and virtual infrastructure a feature parity between the 2 apps would really be needed - or everything must be done for the lowest common denominator (i.e. if some feature is only for `dcim` but not `virtualization` then it could not be used).

Hey @bluikko, I can look back through when I get a chance.
Off the top of my head, I don't remember seeing anything that would cause a virtual device to operate differently than a physical one.

k-y:
I've got the site object getting populated correctly into hostvars and see a single prefix in site, however, only the 1st prefix for each site seems to be getting populated.  All of our sites have prefix_count > 1 but I don't see the rest of the prefixes.

e.g.
```
        "site": {                                                                 
...
            "prefix": {    
                "created": "2021-09-16",
...
            },
            "prefix_count": 23,
```

rodvand:
@k-y Could you open an issue with enough data to reproduce?

Eric-Sim:
I've spoken with them and I'll open the ticket.

...and fix it. :)

PR #665, opened 2021-11-30
Replace FileNotFoundError by OSError to stay compatible with Python 2

### LINKED ISSUE #661: Python 2 compatibility issue with FileNotFoundError
##### ISSUE TYPE
 - Bug Report

##### SOFTWARE VERSIONS
- CentOS 7.9
- Python 2.7.5

###### Ansible:
- Ansible 2.9

###### NetBox:
- NetBox 3.0.11

###### Collection:
- 3.1.1

##### SUMMARY
Lookup plugin nb_lookup.py is not compatible with Python 2

##### STEPS TO REPRODUCE
Run any playbook containing a query to the nb_lookup plugin with e.g. the "api_endpoint" parameter missing.

```yaml
- name: Test NetBox DNS lookup
  hosts: localhost
  tasks:

    - name: Look up Devices
      debug:
          msg: "{{ query('netbox.netbox.nb_lookup',
                         'devices',
                         token='<redacted>') }}"
```

##### EXPECTED RESULTS
The lookup returns a proper error message.

##### ACTUAL RESULTS
```
PLAY [Test NetBox DNS lookup] *************************************************************************************

TASK [Gathering Facts] ********************************************************************************************
ok: [localhost]

TASK [Look up Devices] ********************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'netbox.netbox.nb_lookup'. Error was a <type 'exceptions.NameError'>, original message: global name 'FileNotFoundError' is not defined"}

PLAY RECAP ********************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0 
```

The reason is that PR #52 for issue #49 uses FileNotFoundError to catch the exception. This exception is only supported in Python 3.

peteeckel:
fixes #661

This still does not provide a sensible error message for a missing API endpoint as outlined in the steps to reproduce #661, but at least it does no longer try to raise an exception not available on Python 2.

### DIFF
--- a/plugins/lookup/nb_lookup.py
+++ b/plugins/lookup/nb_lookup.py
@@ -380,7 +380,7 @@ class LookupModule(LookupBase):
                 private_key_file=netbox_private_key_file,
             )
             netbox.http_session = session
-        except FileNotFoundError:
+        except OSError:
             raise AnsibleError(
                 "%s cannot be found. Please make sure file exists."
                 % netbox_private_key_file


sc68cal (changes requested):
Python 2 is end-of-life as of Jan 1 2020

peteeckel:
I know. 

On the other hand
* the compatibility list of the Ansible modules still lists Python 2
* RHEL as the main sponsor of Ansible still provides Ansible 2.9 on EL7, which still runs on Python 2

Especially the latter argument is probably a valid cause for keeping Python 2 compatibility until EL 7 goes EOL in 2024.

And no, I don't like it either.

PR #673, opened 2021-12-15
adding pynetbox check and wireless endpoints for netbox v3.1

### LINKED ISSUE #670: Add nb_lookup endpoints for NetBox 3.1 support
##### ISSUE TYPE
- Feature

##### SUMMARY
Add endpoints defined in NetBox 3.1 to support lookups with 3.1 using the ansible collection.

netbox.wireless.wireless_lan_groups
netbox.wireless.wireless_lan_groups
netbox.wireless.wireless_lan
netbox.wireless.wireless_links

These endpoints are supported in pynetbox 6.4.0.

ThomasADavis:
This fixes #670 

Adds a check for pynetbox v6.4.0.  If the current pynetbox is too old, simply says so when run with a '-v'.

### DIFF
--- a/plugins/lookup/nb_lookup.py
+++ b/plugins/lookup/nb_lookup.py
@@ -256,19 +256,25 @@ def get_endpoint(netbox, term):
         "webhooks": {"endpoint": netbox.extras.webhooks},
     }
 
-    major, minor, patch = [int(x, 10) for x in pynetbox.__version__.split('.')]
+    major, minor, patch = [int(x, 10) for x in pynetbox.__version__.split(".")]
 
     if major >= 6 and minor >= 4 and patch >= 0:
-        netbox_endpoint_map["wireless-lan-groups"] = {"endpoint": netbox.wireless.wireless_lan_groups}
-        netbox_endpoint_map["wireless-lan-groups"] = {"endpoint": netbox.wireless.wireless_lan_groups}
+        netbox_endpoint_map["wireless-lan-groups"] = {
+            "endpoint": netbox.wireless.wireless_lan_groups
+        }
+        netbox_endpoint_map["wireless-lan-groups"] = {
+            "endpoint": netbox.wireless.wireless_lan_groups
+        }
         netbox_endpoint_map["wireless-lan"] = {"endpoint": netbox.wireless.wireless_lan}
-        netbox_endpoint_map["wireless-links"] = {"endpoint": netbox.wireless.wireless_links}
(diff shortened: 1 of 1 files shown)


plugins/lookup/nb_lookup.py
```diff
@@ -256,6 +256,26 @@ def get_endpoint(netbox, term):
         "webhooks": {"endpoint": netbox.extras.webhooks},
     }
 
+    major, minor, patch = [int(x, 10) for x in pynetbox.__version__.split(".")]
```
rodvand, line 259: Not a big deal, but I think you can do it like this as well: 
`major, minor, patch = map(int, pynetbox.__version__.split("."))`

ThomasADavis: Put me down as knows enough python to be dangerous, but been around too long to keep track of all the changes..  looking at map vs. for loop, map is faster, so map wins.

plugins/lookup/nb_lookup.py
```diff
@@ -256,6 +256,26 @@ def get_endpoint(netbox, term):
...
+        }
+        netbox_endpoint_map["wireless-lan-groups"] = {
+            "endpoint": netbox.wireless.wireless_lan_groups
+        }
+        netbox_endpoint_map["wireless-lan"] = {"endpoint": netbox.wireless.wireless_lan}
+        netbox_endpoint_map["wireless-links"] = {
+            "endpoint": netbox.wireless.wireless_links
+        }
+
+    else:
```
sc68cal, line 273: I don't think we want this to get printed all the time

sc68cal (changes requested):

plugins/lookup/nb_lookup.py
```diff
@@ -256,6 +256,26 @@ def get_endpoint(netbox, term):
...
+        Display().v(
```
sc68cal, line 274: `-v` is going to end up printing this in everyone's playbooks. This is not the right verbosity level.

sc68cal (changes requested):

plugins/lookup/nb_lookup.py
ThomasADavis: Please advise on what the right verbosity level you believe it should be.  I've changed it to only complain when you try to use the wireless-* lookups.

sc68cal (approved):

plugins/lookup/nb_lookup.py
sc68cal: That sounds good to me

PR #682, opened 2021-12-29
Close #464 - Add parent_inventory_item to module

### LINKED ISSUE #464: Missing 'parent' and 'label' paramaters in netbox.netbox.netbox_inventory_item
Hi all! 
I miss for the parameters 'parent' and 'label', because it helps to model parent/child relations between let's say module, slot of module and transceiver. 

`Unsupported parameters for (netbox.netbox.netbox_inventory_item) module: parent found in data. Supported parameters include: asset_tag, description, device, discovered, manufacturer, name, part_id, serial, tags`

Is it possible to add these parameters?

rodvand:


bluikko:
This is a great addition.

A few questions/suggestions though:
- It is not very clear from the documentation block how exactly is `parent_inventory_item` used. Maybe a dict of `name` and `device`?
- If `parent_inventory_item` would be added to the examples block then it might be more clear how exactly to use it.

PR #701, opened 2022-01-20
inventory: Add an option bypass querying racks

bvergnaud:
Hello 👋🏻 

This PR adds an option to bypass querying racks and related objects in the inventory plugin.

The motivation behind it is that in our NetBox, we have a very large amount of racks registered and, with pagination, it takes a good 40~45 seconds to gather that information, for a very limited benefit as we don't use it at all. Of course it can be mitigated by using caching but even then, it still happens occasionally when the cache expires / needs to be refreshed.

The option, by default, is set to not change anything in the current behaviour.

For the record, I've had some difficulty updating the tests. Most notably, I never managed to make the `hacking/update_test_inventories.sh` script work, so I've had to do some manual work to get to the current point and I could very well have messed up. I've also not found how to update the sphinx docs, I've seen the `Makefile` but I'm unsure what options to pass to sphinx, and the `CONTRIBUTING.md` document does not mention it at all. It might be trivial but I'm unfamiliar with sphinx.

If someone is willing to point me in the right direction for those two points, I'll happily include a commit to freshen up the relevant sections in `CONTRIBUTING.md`, or add them if need be.

Cheers.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -133,7 +133,7 @@ DOCUMENTATION = """
             version_added: "0.2.1"
         group_by:
             description:
-                - Keys used to create groups. The I(plurals) option controls which of these are valid.
+                - Keys used to create groups. The I(plurals) and I(racks) options control which of these are valid.
                 - I(rack_group) is supported on NetBox versions 2.10 or lower only
                 - I(location) is supported on NetBox versions 2.11 or higher only
             type: list
@@ -214,6 +214,12 @@ DOCUMENTATION = """
             description: List of custom ansible host vars to create from the device object fetched from NetBox
             default: {}
             type: dict
+        racks:
+            description:
+                - If False, skip querying the racks for information, which can be slow with great amounts of racks.
+                - The choices of I(group_by) will be changed by this option.
+            type: boolean
+            default: True
 """
 
 EXAMPLES = """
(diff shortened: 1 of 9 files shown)


rodvand:
Any input on this @nahun or @sc68cal?

nahun:
I would change the option name to just `racks` instead of `query_racks`. The other options like this are just `interfaces` or `dns_name`.

Otherwise it looks good.

The `update_test_inventories.sh` script never worked for me either without manual changes. I'm pretty unfamiliar with sphinx too. The testing framework for the inventory plugin needs some work and I think should be overhauled to function how the [AWS collection handles it](https://github.com/ansible-collections/amazon.aws/tree/main/tests/integration/targets/inventory_aws_ec2). But I'm getting too far ahead...

The additions to the testing look good.

sc68cal:
The code looked good to me when I looked at it last week, but I was unsure about how to proceed with the issue the author mentioned about the `update_test_inventories.sh` script.

sc68cal (approved):

sc68cal:
The only question I have, is if for consistency we should have the default behavior be that rack lookups are not enabled, similar to how `prefixes` are not enabled by default? Just throwing it out there for discussion.

nahun:
Currently, rack info is always returned since there is no option for it. So by default we want the new option to be enabled so that we don't break user's environments when they upgrade. Consistency would be nice for sure, but don't think we should change the default behavior.

rodvand:
Yes, we shouldn't change any default behaviour. @bvergnaud Could you look at the input from @nahun and I'll merge this as soon as it's sorted?

Yes, the update test inventories script is not working out-of-the-box. I'll create an issue so that I can update the documentation work.

bvergnaud:
The rename from `query_racks` to `racks` is done. Thanks for your feedbacks. :)

Issue #732, opened 2022-02-09
[Bug]: Inventory Module will only load 1 prefix when multiple are associated to a site


Eric-Sim:
### Ansible NetBox Collection version

Development branch

### Ansible version

```bash
ansible [core 2.11.4]
  config file = None
  configured module search path = ['/Users/eric.sim/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/eric.sim/.pyenv/versions/3.8.9/envs/ansible-test/lib/python3.8/site-packages/ansible
  ansible collection location = /Users/eric.sim/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/eric.sim/.pyenv/versions/ansible-test/bin/ansible
  python version = 3.8.9 (default, Aug 25 2021, 11:07:09) [Clang 12.0.5 (clang-1205.0.22.11)]
  jinja version = 2.11.2
  libyaml = False
```

### NetBox version

2.11.8

### Python version

3.8

### Steps to Reproduce

Populate Netbox with at least one site having more than 1 prefix associated to it.
Load inventory from Netbox using netbox inventory plugin (must specify the "site_data" and "prefixes" options as True.)

### Expected Behavior

Expected behavior is that all prefixes associated to a site are included in the Ansible inventory.

### Observed Behavior

Only 1 prefix for each site is inserted into the site data in Ansible Inventory.

Eric-Sim:
I will submit a PR to fix this (likely next week)
Assign bug to me if you wish.

rodvand:
Have you had a chance to look at this @Eric-Sim ? Wondering if I should do a new release of the collection this week, could be nice to include this fix if possible.

Eric-Sim:
I've identified the issue and can have the solution tested tomorrow.

PR #742, opened 2022-02-23
Multiple prefix per site - BUG #732

Eric-Sim:
## New Behavior

...

...

...

None

Inventory module now properly includes the full list of prefixes assigned to a site.

## Double Check

* [ X ] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netbox-community/ansible_modules/blob/devel/CONTRIBUTING.md).
* [ X ] I have explained my PR according to the information in the comments or in a linked issue.
* [ X ] My PR targets the `devel` branch.

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -648,10 +648,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             site = self.sites_lookup[host["site"]["id"]]
             if (
                 self.prefixes
-            ):  # If prefixes have been pulled, attach prefix to its assigned site
-                prefix_id = self.prefixes_sites_lookup[site["id"]]
-                prefix = self.prefixes_lookup[prefix_id]
-                site["prefix"] = prefix
+            ):  # If prefixes have been pulled, attach prefix list to its assigned site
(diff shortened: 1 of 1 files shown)


Eric-Sim:
I think that 3.5.1 (about a month ago) is the only release that includes the site data and prefixes, so people who are using that version definitely could have seen the singular "prefix" data in their inventory.

Any potential user who has built a workflow around having the singular "prefix" attribute nested within site data is going to  be disrupted by this fix. 

That being said, my thinking was that it would actually be better to have those users error out by trying to grab a key that no longer exists (which they would investigate and quickly figure out). The alternative, it that they continue to being able to access the same key, but be returned a different data structure. (This way may be more difficult for someone to notice and troubleshoot if they don't pay close attention to release notes)

rodvand:
I'm thinking it could go in as a bug fix... the functionality has not worked the way it's supposed to be so anyone currently using it is getting a half-working solution.

Eric-Sim:
I'd agree with calling it a bug fix.
I'll defer to you guys on how you'd like this to be dealt with though.

rodvand (approved):

sc68cal:
I think we should call it out in the release notes that the key has changed from `prefix` to `prefixes` - just to close the loop.

PR #753, opened 2022-03-13
Fixes #752 - Add meta information for EE

### LINKED ISSUE #752: [Feature]: Collection-level Metadata for better execution-environment handling
### Change Type

Addition

### Area

Other

### Proposed Changes

Please add  and maintain the python requirements (modules: pynetbox / packages) of the collection in its Collection-level Metadata so that people can more easily use the collection in execution-environments when using it with ansible-tower or awx.

The Collection-level Metadata makes it more easy to include the collection in execution-environments, since the python dependencies will be automatically installed, when building the execution-environment. 

Furthermore you could track the dependencies in context of a release.

If you need further information about the Collection-level Metadata here is some Documentation from the ansible-builder Project [Documentation](https://ansible-builder.readthedocs.io/en/stable/collection_metadata.html).

Thanks in  advance

rodvand:
## New Behavior
```
(builder) ➜  custom-awx-ee git:(main) ✗ ansible-builder introspect /Users/rodvand/collections/
# Dependency data for /Users/rodvand/collections/
---
python:
  netbox.netbox:
  - pynetbox
  - packages
system: {}
```

```
(builder) ➜  custom-awx-ee git:(main) ✗ ansible-builder introspect /Users/rodvand/collections/
# Dependency data for /Users/rodvand/collections/
---
python: {}
system: {}
```
## Double Check

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+pynetbox
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 pynetbox
+packages
--- a/meta/execution-environment.yml
+++ b/meta/execution-environment.yml
@@ -1,0 +1,5 @@
+---
+version: 1
+
+dependencies:
+  python: requirements.txt


sc68cal (approved):

requirements.txt
```diff
@@ -0,0 +1,2 @@
+pynetbox
+packages
```
sc68cal, line 2: Is there a way to limit this to when Ansible versions < 2.10 is installed? I did a brief look at https://peps.python.org/pep-0508/#environment-markers but don't have experience with it to make a useful suggestion for how to implement it

sc68cal:
I'm actually digging into this more, and [packages](https://pypi.org/project/packages/) doesn't appear to do anything. I extracted both the source tarball and the wheel and it's an empty python file. 

So, do we really need `packages`?

Or was it a typo, and @FragmentedPacket actually meant `packaging`? https://github.com/pypa/packaging

sc68cal (changes requested):
Packages requirement may be invalid

nwerker:
As described in #752 I could sucessfully test the functionality, thanks again! 

Maybe you could also include the minimum version like stated in your Readme! This could be helpful when people work with cached / selfhosted pypi servers and maybe don't have the desired versions available.

This would mean for the requirements.txt something like:
```
pynetbox >=5.0.4
```

Also if in the future, for whatever reason, a certain new collection version shouldn't be working with the newest pynetbox version you could also limit the maxiumum version like:
```
pynetbox >=5.0.4, <=9.9.9
```

sc68cal:
I don't think we'd pin pynetbox at specific versions since our intent is to be closely coupled to that library

sc68cal:
@rodvand ca04151 looks good, but can you also update the README? thanks

PR #755, opened 2022-03-29
add site_group to netbox_inventory

### LINKED ISSUE #666: [Feature]: Inventory Plugin site_group
##### ISSUE TYPE
 - Feature Idea

###### Ansible:
2.11.6

###### NetBox:
3.0.11

###### Collection:
3.4.0

##### SUMMARY
Add Inventory Plugin site_group

##### EXPECTED RESULTS

Be able to leverage site_group in netbox_inventory for functions like query_filters and group_by.

ryanmerolle:
## Related Issue

Closes #666 

## New Behavior

- Add site_group to netbox_inventory plugin
- Update tests related to site_group

site_group is not aa feature within the netbox_inventory plugin

Benefit is to have feature parity with NetBox.

TBD

Add site_group to netbox_inventory plugin.

## Double Check


ryanmerolle:
@rodvand please remind me of the doc update process if you could.

ryanmerolle:
While digging into site_group, I noticed group (site_group) was applied to some modules like devices.  Is this not confusing given the model is site_group but group seems to refer to the site_group model?  I can address this in another PR if you agree it should be standardized @rodvand, but this will be a breaking change for anyone using site_group with devices.

ryanmerolle:
Ha, this pull definitely needs a swish when it's merged.

rodvand:
> While digging into site_group, I noticed group (site_group) was applied to some modules like devices. Is this not confusing given the model is site_group but group seems to refer to the site_group model? I can address this in another PR if you agree it should be standardized @rodvand, but this will be a breaking change for anyone using site_group with devices.

Could it be that way since the variable name in the API on a site to reference a site group is `group`. So instead of transforming it to `site_group` the default has just been used? If we are to change it (and release a new major version) I think we need to gather up some more breaking changes to make it "worthwhile".

rodvand:
> @rodvand please remind me of the doc update process if you could.

Don't think this needs a big doc update. It's just another option in the group_by. I will add a change log entry when making a new release.

ryanmerolle:
> > While digging into site_group, I noticed group (site_group) was applied to some modules like devices. Is this not confusing given the model is site_group but group seems to refer to the site_group model? I can address this in another PR if you agree it should be standardized @rodvand, but this will be a breaking change for anyone using site_group with devices.
> 
> Could it be that way since the variable name in the API on a site to reference a site group is `group`. So instead of transforming it to `site_group` the default has just been used? If we are to change it (and release a new major version) I think we need to gather up some more breaking changes to make it "worthwhile".

I am inclined to leave the reference of group as is in the other modules.

ryanmerolle:
Since this just group_by and the hostvar for key for site_groups is being added, and the docs use the python modules themselves to document.  I think this is in a state that can be merged.  Final thoughts @rodvand ?

rodvand (approved):

PR #759, opened 2022-03-30
replace NetBox 3.2 testing

### LINKED ISSUE #756: [Housekeeping]: 3.2 support
### NetBox version

v3.2.0

### Feature type

Change to existing Module

### Proposed functionality

- [ ] Update collection to support NetBox 3.2 (should be minimal work if any)
- [ ] Drop NetBox v2.11 testing and replace with NetBox v3.2 testing

### Use case

We only test against the current and last 2 major releases of NetBox.

### External dependencies

_No response_

ryanmerolle:
## Related Issue

Closes #756 

## New Behavior

Integration testing dropped for 2.11 and replaced with 3.2

Testing supports currently 2.11-3.1

Add tests for 3.2

N/A

Integration testing dropped for 2.11 and replaced with 3.2

## Double Check


sc68cal:
The only concern I have is dropping support for 2.11 which was only released last year. I would like to keep support around for a little longer.

ryanmerolle:
> The only concern I have is dropping support for 2.11 which was only released last year. I would like to keep support around for a little longer.

I mean its a fair point, but the project testing has always tested against:

- The last 3 major NetBox releases.
- The last 3 major python releases.  
- The last non-RC ansible

I was really only reflecting that.

The first release of NetBox 2.11.0 was 2021-04-16.  The last release of NetBox 2.11.11 was 2021-08-23.

I can add 2.11 back if you feel strongly.

sc68cal:
I do feel strongly, since this is the first NetBox version where they have had a major API change that required bumping the semantic version. I think we will need to keep it around for a while, since my org has not upgraded from 2.11 to 3.x yet, so I can't imagine that I'm the only one.

ryanmerolle:
> I do feel strongly, since this is the first NetBox version where they have had a major API change that required bumping the semantic version. I think we will need to keep it around for a while, since my org has not upgraded from 2.11 to 3.x yet, so I can't imagine that I'm the only one.

Good enough of a reason for me.

rodvand:
> I do feel strongly, since this is the first NetBox version where they have had a major API change that required bumping the semantic version. I think we will need to keep it around for a while, since my org has not upgraded from 2.11 to 3.x yet, so I can't imagine that I'm the only one.

I'm not sure the API changes were a reason to bump the versioning, I think it was more related to the massive UI overhaul.

sc68cal:
> I'm not sure the API changes were a reason to bump the versioning, I think it was more related to the massive UI overhaul.

There were API changes, I think the removal of endpoints was what caused the semver bump

https://netbox.readthedocs.io/en/stable/release-notes/version-3.0/

```
    Removed all endpoints related to the secrets functionality:
        /api/secrets/generate-rsa-key-pair/
        /api/secrets/get-session-key/
        /api/secrets/secrets/
        /api/secrets/secret-roles/
    Removed the following "connections" endpoints:
        /api/dcim/console-connections/
        /api/dcim/power-connections/
        /api/dcim/interface-connections/
    Added the /api/ipam/ip-ranges/ endpoint
    Added the /api/users/tokens/ endpoint
        The provision/ child endpoint can be used to provision new REST API tokens by supplying a valid username and password
    dcim.Cable
        length is now a decimal value
    dcim.Device
        Removed the display_name attribute (use display instead)
    dcim.DeviceType
        Removed the display_name attribute (use display instead)
    dcim.FrontPort
        Added color field
    dcim.FrontPortTemplate
        Added color field
    dcim.Rack
        Removed the display_name attribute (use display instead)
    dcim.RearPort
        Added color field
    dcim.RearPortTemplate
        Added color field
    dcim.Site
        latitude and longitude are now decimal fields rather than strings
    extras.ContentType
        Removed the display_name attribute (use display instead)
    ipam.Prefix
        Added the mark_utilized boolean field
    ipam.VLAN
        Removed the display_name attribute (use display instead)
    ipam.VRF
        Removed the display_name attribute (use display instead)
    virtualization.VirtualMachine
        vcpus is now a decimal field rather than a string
```

Issue #799, opened 2022-06-15
[Feature]:  Support Ansible module defaults groups


jensrudolf:
### NetBox version

v3.1.6

### Feature type

Change to existing Module

### Proposed functionality

This collection should support module defaults groups as documented here: https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html#module-defaults-groups

### Use case

This enables specifying default parameters as e.g. ``url`` or ``token`` only once for all used modules and reduces boilerplate code.

PR #800, opened 2022-06-15
Add action_group to enable module defaults groups

jensrudolf:
## Related Issue

#799

## New Behavior

This enables support for the module defaults groups feature introduced in ansible-core 2.12. Using module defaults groups one can provide default values (e.g. token, url, etc.) for a group of modules instead of having to specify them for every used module. For more details see: https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html#module-defaults-groups

This commit adds all modules available in this collection to an action group called 'netbox'. This change is backwards compatible. Older versions of Ansible just ignore this.

Currently, user has to specify module defaults (e.g. url, token) for every module in use.

This change potentially reduces the amount of boilerplate code using this collection as required parameters like e.g. ``url`` or ``token`` need to be provided once and can be specified in a central place.

Might add a hint to the documentation about the usage of this feature.

Support module defaults groups

## Double Check

### DIFF
--- a/docs/getting_started/how-to-use/advanced.rst
+++ b/docs/getting_started/how-to-use/advanced.rst
@@ -299,3 +299,35 @@ IP address to be within **Test VRF 2**.
 We can now see that the IP address is now within VRF with ID 2.
 
 Hopefully this shines some light on this useful feature to allow you, as the user, to define your specific needs for finding a unique object within your NetBox instance.
+
+
+Using Module defaults groups
+--------------------------------------------
+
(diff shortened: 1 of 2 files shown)


sc68cal:
This looks good, I'd just like to see the documentation that makes this work, since I am not familiar with it myself, and the quicker we update the docs for it, the quicker everyone stops putting the `url` and `token` in every netbox module call

rodvand (approved):

sc68cal:
@jensrudolf awesome work, thank you for the contribution!

PR #806, opened 2022-06-27
Allow netbox api access token to be templated

TWitzenrath:
## New Behavior

You can use jinja templating for the API access token in the nb_inventory 
config.

The only possibility not to have the token in clear text in the config was to
use ansible vault.

This makes it possible to use a lookup() to retrieve the token, i.e. from a 
password-manager.

Idea is stolen from the hcloud inventory.

token:  NetBox API token to be able to read against NetBox. This may be templated with jinja i.e. to use lookup().

inventory token: can now use jinja templating i.e. to use lookup()

## Double Check

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -1853,7 +1853,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         self.use_cache = cache
 
         # NetBox access
-        token = self.get_option("token")
+        token = self.templar.template(self.get_option("token"), fail_on_undefined=False)
         # Handle extra "/" from api_endpoint configuration and trim if necessary, see PR#49943
         self.api_endpoint = self.get_option("api_endpoint").strip("/")
         self.timeout = self.get_option("timeout")


bk2zsto:
omgz! Just wasted a day trying to figure out how to do this. I can confirm it works. Thank you!

rodvand:
Hi @TWitzenrath ! Thanks for the PR, and sorry for the slow response. Could you elaborate a bit on how this would be used? As in an example we can provide in the documentation.

TWitzenrath:
@rodvand sure:
right now, you can put 
``
api_endpoint: "https://netbox.example.com/"
token: !vault | [...]
``
or
``
api_endpoint: "https://netbox.example.com/"
token: cleartexttoken
``
in your inventory yaml file. The ansible-vault encrypted token will be replaced by ansible, but anything else will be left as is, so if you put i.e. "{{ my_token }}" as a token, nb_inventory will NOT send the content of my_token to netbox as API-token, but will send the literal string "{{ my_token }}". 

With the change merged, you can put i.e.

``
api_endpoint: "https://netbox.example.com/"
token: "{{ lookup('community.hashi_vault.hashi_vault','secret=[...path...]/netbox:api_token')}}"
``
to retrieve the netbox api token from the hashi_vault plugin or any other lookup/jinja template.

PR #859, opened 2022-11-02
Fix two idempotency issues when updating VMs.

### LINKED ISSUE #823: [Bug]: Swapped behaviour for vCPUs on VMs between pre and post 3.0.0

v3.7.1

### Ansible version

```bash
ansible [core 2.13.1]
  config file = /home/bvergnaud/git/scw/platform/vm-containers/iaas-playbooks/ansible.cfg
  configured module search path = ['/home/bvergnaud/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/bvergnaud/.local/share/virtualenvs/iaas-playbooks-3YYUb2kg/lib/python3.8/site-packages/ansible
  ansible collection location = /home/bvergnaud/git/scw/platform/vm-containers/iaas-playbooks/.collections
  executable location = /home/bvergnaud/.local/share/virtualenvs/iaas-playbooks-3YYUb2kg/bin/ansible
  python version = 3.8.13 (default, Apr  4 2022, 10:14:53) [GCC 11.2.0]
  jinja version = 3.1.2
  libyaml = True
```

### NetBox version

v2.10.10

### Python version

3.8

### Steps to Reproduce

Running a simple `netbox.netbox.netbox_virtual_machine` task twice, with the vcpus parameter set:
```yml
- name: "Create/delete VM on NetBox"
  netbox.netbox.netbox_virtual_machine:
    netbox_url: "{{ netbox_url }}"
    netbox_token: "{{ netbox_token }}"
    data:
      name: test-vm
      vcpus: 1
    state: present
  delegate_to: localhost
```

### Expected Behavior

The task should be idempotent and report status=ok with no changes.

### Observed Behavior

On the second run, it produces a diff:
```diff
--- before
+++ after
@@ -1 +1 @@
-vcpus: 1
+vcpus: '1.00'

changed: [lab1dc2-inf-dev-hv16 -> localhost] => changed=true
  msg: virtual_machine test-vm updated
  virtual_machine:
    comments: ''
    config_context: {}
    created: '2022-09-06'
    custom_fields:
      vmid: null
    id: 9094
    last_updated: '2022-09-06T14:00:18.040069Z'
    local_context_data: null
    name: test-vm
    status: active
    url: https://netbox-staging.internal.scaleway.com/api/virtualization/virtual-machines/9094/
    vcpus: 1.0
```

Note:
* Output filtered manually from testing with more parameters, in case anything looks odd.
* The same behaviour holds true if we cast the value of `vcpus` as int or float.

bvergnaud:
## Related Issue

#823

## New Behavior

Using `netbox.netbox.netbox_virtual_machine` to update the vCPU count of a VM in netbox is now idempotent from version 2.10 onward.
Using `netbox.netbox.netbox_virtual_machine` to update a VM in netbox that has custom_fields with no values is also idempotent from version 2.10 onward.

Plus a couple of pretty minor changes (that can be removed easily if undesired):
* Method `_version_check_greater` now returns a type consistent value (i.e. False instead of None by default).
* Fix for a typo in an error message.

In the current behaviour, I'd get a float value replacing my integer value on the `vcpus` field every time (netbox 2.10).
And if I had a `custom_fields` with value `None`, the module would systematically return a `state: changed` with an empty diff (netbox 2.10 and 3.2).

To do that, I dropped the use of a formatted string and instead use the same types as pynetbox, i.e. integer / float, depending on the situation. The user can use either type and it'll work fine, so long as pynetbox and the targeted netbox are able to handle it. Otherwise, an error is raised, e.g. `pynetbox.core.query.RequestError: The request failed with code 400 Bad Request: {'vcpus': ['Ensure that there are no more than 6 digits in total.']}`

I manually tested on 2.10 and 3.2 instances, with different `vcpus` values (2, 2.0, 2.1, 2.10, 2.12), and I get a consistent, idempotent behaviour.

Additionally, there was a piece of code that would always generate diffs if the original or updated objects contained a custom field with a value of None.
I chose to replicate the existing behaviour on the updated object rather than delete the loop over the original object. But admittedly, I do not understand the purpose of this piece of code. If anyone has knowledge about this, I'd be happy to learn more.

Existing tests pass, and I've also added an idempotence test on the execution of the `netbox_virtual_machine` module.

Fix #823: Fix idempotence of netbox_virtual_machine module.

## Double Check

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -1298,7 +1298,7 @@ class NetboxModule(object):
                         data_after[key] = updated_obj[key]
                 except KeyError:
                     if key == "form_factor":
-                        msg = "form_factor is not valid for NetBox 2.7 onword. Please use the type key instead."
+                        msg = "form_factor is not valid for NetBox 2.7 onward. Please use the type key instead."
                     else:
                         msg = (
                             "%s does not exist on existing object. Check to make sure valid field."
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -676,6 +676,8 @@ class NetboxModule(object):
         elif g_major == l_major and g_minor > l_minor:
             return True
 
+        return False
+
     def _connect_netbox_api(self, url, token, ssl_verify, cert):
         try:
             session = requests.Session()
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -1257,12 +1257,28 @@ class NetboxModule(object):
         """
         serialized_nb_obj = self.nb_object.serialize()
         updated_obj = serialized_nb_obj.copy()
+
+        if serialized_nb_obj.get("custom_fields"):
+            serialized_nb_obj["custom_fields"] = {
+                key: value
+                for key, value in serialized_nb_obj["custom_fields"].items()
+                if value is not None
+            }
+
+        if updated_obj.get("custom_fields"):
+            updated_obj["custom_fields"] = {
+                key: value
+                for key, value in updated_obj["custom_fields"].items()
+                if value is not None
+            }
+
         updated_obj.update(data)
+
         if serialized_nb_obj.get("tags") and data.get("tags"):
             serialized_nb_obj["tags"] = set(serialized_nb_obj["tags"])
             updated_obj["tags"] = set(data["tags"])
 
-        # Ensure idempotency for site and virtual machine on version pre-3.0
+        # Ensure idempotency for site on older netbox versions
         version_pre_30 = self._version_check_greater("3.0", self.version)
         if (
             serialized_nb_obj.get("latitude")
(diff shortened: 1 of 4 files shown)


sc68cal (changes requested):
I don't get why you added the `@staticmethod` decorator to `_version_check_greater` yet all the callers to the function are still via `self._version_check_greater` instead of calling the class.

I think that change should be dropped from this PR.

bvergnaud:
> I don't get why you added the `@staticmethod` decorator to `_version_check_greater` yet all the callers to the function are still via `self._version_check_greater` instead of calling the class.
> 
> I think that change should be dropped from this PR.

Calling the class instead of self would merely spawn a new instance of that class. Since we're already in the class, that would be quite unnecessary, since self is also an instance of the class. A static method does not HAVE to be called via a new instance of the class.

But granted, this is not necessary for this PR so if you don't feel comfortable merging it, I don't mind dropping the commit.

bvergnaud:
Update:

* Tests pass, and I've added a new idempotence test for the virtual_machine module.
* Fixed another issue I didn't anticipate: if a VM has a custom_field with a value of None, it would systematically cause a `status: changed`, even with an empty diff. I'm not sure the way I fixed it is the best way. [Here](https://github.com/netbox-community/ansible_modules/pull/859/files#diff-52074cf41ca073c05dda77695b9c5f1fb5f6920a0cfc3b2cb1d78d16dbad7965R1288), there's a loop that removes such custom fields from the original, serialized object. I added the same behaviour on the updated object to make the comparison fair. But I don't get the purpose of this behaviour. It could be cleaner to instead not touch either object before the comparison, but I didn't want to risk silently breaking something.

As is, the PR is good to go, if there are no further comments / changes requests.

sc68cal (approved):

PR #896, opened 2022-12-02
Add custom headers to inventory request

bloomcake:
## New Behavior

* Add custom headers to inventory request

New `headers` are appended to `self.headers` to be able to add own headers.

...

...

Add custom headers to inventory request

## Double Check

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -232,6 +232,10 @@ DOCUMENTATION = """
             type: boolean
             default: True
             version_added: "3.6.0"
+        headers:
+            description: Dictionary of headers to be passed to the NetBox API.
+            type: dict
+            default: {}
 """
 
 EXAMPLES = """
@@ -249,6 +253,8 @@ query_filters:
 device_query_filters:
   - has_primary_ip: 'true'
   - tenant__n: internal
+headers:
+  Cookie: "{{ auth_cookie }}"
 
 # has_primary_ip is a useful way to filter out patch panels and other passive devices
 # Adding '__n' to a field searches for the negation of the value.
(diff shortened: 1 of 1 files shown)


sc68cal (changes requested):
This should be managed via the standard `HTTP_PROXY` environment variable - https://docs.ansible.com/ansible/2.4/playbooks_environment.html

sc68cal:
pynetbox under the hood uses `requests` - which honors the standard `HTTP_PROXY` environment variable and others

https://requests.readthedocs.io/en/latest/user/advanced/?highlight=HTTP_PROXY#proxies

sc68cal:
Basically, I don't want us to have code in our repository that deals with however you've deployed netbox in your environment. There are generally accepted practices for setting up proxies, we have a couple of users who have reported their usage of netbox behind a proxy. You should configure an HTTP(s) proxy, instead of adding code that sets a cookie and make everyone else who ever uses this code support that.

clinta:
Sadly found this issue trying to figure out how to use this plugin with netbox behind Cloudflare zero trust. I'm not sure how an outbound http proxy is supposed to replace the ability to add custom headers.

sc68cal:
See the linked issue 895, user was using a custom proxy solution which required a custom header to be set in order to work with their proxy setup. I don't wish for us to carry the burden of maintaining how people set up proxies between the various NetBox components (both API clients and the netbox server), since it also has been a source of bug reports where we don't really have a way of fixing it since it's unique to each user's environment.

clinta:
I can understand that for a complex change. But this change is literally one line of (of non-documentation). Also this functionality is explicitly supported by pynetbox, and [documented](https://pynetbox.readthedocs.io/en/latest/advanced.html#headers). The implementation does not have any specific logic for dealing with specific proxies, it's just the generic ability to add any headers the user needs. 

I've rebased this PR to run my own fork, and added the same functionality to the lookup module. I'd ask you to reconsider and look at what a small change this is.

https://github.com/netbox-community/ansible_modules/compare/master..clinta:custom_headers

sc68cal:
OK, that is a convincing argument to me. Open a PR. I'll :+1:

PR #900, opened 2022-12-09
Ensure that packaging is installed

sc68cal:


sc68cal:
I sort of low key hate black.

rodvand (approved):

Issue #925, opened 2023-01-11
[Bug]: 


jhofmueller:
### Ansible NetBox Collection version

v3.10.0

### Ansible version

```bash
ansible [core 2.12.8]
  config file = /home/jogi/git/dcim/netbox-dev/ansible.cfg
  configured module search path = ['/home/jogi/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/jogi/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0]
  jinja version = 3.0.3
  libyaml = True
```

### NetBox version

v3.3.6

### Python version

3.10

### Steps to Reproduce

Try to create a netbox webhook with a condition like:

  condition: '{"and": [ {"attr": "status.value", "value": "active"} ] }

on ipam.ipaddress.

### Expected Behavior

A webhook with the above specified condition should be created.

### Observed Behavior

An exception occurred during task execution. To see the full traceback, use -vvv. The error was: pynetbox.core.query.RequestError: The request failed with code 400 Bad Request: {'conditions': ["Ruleset must be a dictionary, not <class 'str'>."]}

sc68cal:
That change looks good to me, propose a PR and I'll 👍  it

sc68cal:
Did they change the type for conditions in the NetBox API recently?

PR #926, opened 2023-01-12
Conditions setup fix

tbalg:
## Related Issue

#925

## New Behavior

Conditions setup runs as expected.

Playbook runs without error when `conditions` section is defined.

Fixes the NetBox API's _400 Bad Request_ answer when conditional webhook setup required.
Conditions must be formatted as dictionary.

Type was modified in item description.

netbox_webhook - Fix conditional webhook setup

## Double Check

### DIFF
--- a/plugins/modules/netbox_webhook.py
+++ b/plugins/modules/netbox_webhook.py
@@ -96,7 +96,7 @@ options:
         description:
           - A set of conditions which determine whether the webhook will be generated.
         required: false
-        type: str      
+        type: dict      
       ssl_verification:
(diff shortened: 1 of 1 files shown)


sc68cal:
Would you be willing to add a set of additional tests that exercise this? It appears we do not have any that adds `conditions` 

See

https://github.com/netbox-community/ansible_modules/blob/5646c37681f837672637b0873e9f3ff4fd3c8530/tests/integration/targets/v3.4/tasks/netbox_webhook.yml for example

Issue #878, opened 2022-11-18
[Bug]: nb_lookup incompatible with netbox_secretstore plugin


marinus81:
### Ansible NetBox Collection version

v.3.9.0

### Ansible version

```bash
ansible [core 2.13.6]
  config file = /home/user/ansible/id4-iac/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.8.10 (default, Jun 22 2022, 20:18:18) [GCC 9.4.0]
  jinja version = 3.0.3
  libyaml = True
```

### NetBox version

v3.3.7

### Python version

3.8

### Steps to Reproduce

- Install a version of NetBox which no longer has native secrets feature included (>3.0)
- Install https://github.com/DanSheps/netbox-secretstore in Netbox
Try to fetch secret from NetBox: 
```    - name: Get Login Credentials from Netbox
        tags:
         - debug
        set_fact:
          ansible_user: "{{ query('netbox.netbox.nb_lookup', 'netbox_secretstore', api_filter='role=windows-administrator&virtual_machine='+inventory_hostname, api_endpoint=netbox_host, token=netbox_token, validate_certs=False , key_file='/home/user/.ssh/id_rsa.pem') }}"
```

### Expected Behavior

Lookup plugin uses netbox-secretstore plugin URL to get decrypted secret

### Observed Behavior

``` 
fatal: [windows-2019-test2]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'netbox.netbox.nb_lookup'. Error was a <class 'pynetbox.core.query.RequestError'>, original message: The requested url: https://netbox.domain.at/api/secrets/get-session-key/?preserve_key=True could not be found.. The requested url: https://netbox.domain.at/api/secrets/get-session-key/?preserve_key=True could not be found."}
```

sc68cal:
So `https://netbox.domain.at/api/secrets/get-session-key/?preserve_key=True` 404's - that sounds like an issue with your netbox installation. Please report this issue to them.

sc68cal:
[...]

PR #928, opened 2023-01-24
Remove pre-NetBox 3.3 features

hollow:
## Related Issue

https://github.com/netbox-community/pynetbox/pull/521

Support for secrets has been removed

Support for secrets has been removed

## Double Check


rodvand:
Secrets is actually a pre 3.0 feature so I think it is safe to remove this now. We do however need to mark the option as deprecated in the module, not just remove it.

hollow:
@rodvand I wasn't sure about the required process so I started by simply removing it, since the update to pynetbox 7.x today broke my ansible playbooks 🙈

sc68cal:
This brings up a question, should we be putting constraints on our dependencies, like pynetbox, since they do make API changes that break things?

bluikko:
@sc68cal Definitely!
Even better would be if these 2 projects would coordinate together since they are totally intertwined.

This change has broken everyone's netbox.netbox...

jefvantongerloo:
Following.

Can confirm, pynetbox upgrade broke my calls through nb_lookup modules.
Found a solution in downgrading to 6.6.2.

It would be great to synchronise the pynetbox requirement to assure compatibility with each netbox collection and netbox version release.

ryanmerolle:
releates to #878

PR #930, opened 2023-01-24
Cap pynetbox

sc68cal:
We should probably be capping our requirements to known good versions that we work with.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-pynetbox
+pynetbox<7
 packaging
 pytz


sc68cal:
This PR is meant to generate discussion about how we should proceed. We find ourselves in a situation where we are attempting to support pre NetBox v3.3 _and_ NetBox 3.4 where the cable API has changed. The problem is the pynetbox 7.x series is only compatible with NetBox 3.4 and above - see https://github.com/netbox-community/pynetbox/releases/tag/v7.0.0

I have never been comfortable with our development strategy of just having `devel` and trying to support multiple versions of the NetBox REST API - I think it would be better to have release branches per NetBox REST API release, and now we have a pynetbox release that is not backwards compatible with older REST API versions, I think it might be worth considering, since at least in branches we can manage the dependencies for our libraries and pin them at apropriate versions.

Something to consider.

rfdrake:
pynetbox 7.0.1 also dropped support for private_key_file, which breaks nb_lookup.

I don't have much to contribute to the discussion of how to handle versioning conflicts across the projects, but I'm going to start using pynetbox<7 for now in my requirements.

Issue #946, opened 2023-02-17
[Bug]: netbox_cable errors on existing cables with termination type circuits.circuitterminations


saruter:
### Ansible NetBox Collection version

v3.11.0

### Ansible version

```bash
 config file = /etc/ansible/ansible.cfg
 configured module search path = ['/home/a00135090T1/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
 ansible python module location = /usr/lib/python3.9/site-packages/ansible
 ansible collection location = /home/a00135090T1/.ansible/collections:/usr/share/ansible/collections
 executable location = /usr/bin/ansible
 python version = 3.9.14 (main, Dec  5 2022, 13:41:22) [GCC 8.5.0 20210514 (Red Hat 8.5.0-17)] (/usr/bin/python3.9)
 jinja version = 3.1.2
 libyaml = True

Pynetbox: 7.0.1
```

### NetBox version

v3.4.4

### Python version

3.9

### Steps to Reproduce

Creating a new cable from termination_a_type: dcim.interface to termination_b_type: "circuits.circuittermination" works fine. But running the same playbook again fails with error in module. No Problems on cables with both termination type "dcim.interface".

Simple Task:
```
 tasks:
    - name: Create Cable
      netbox_cable:
        netbox_url: "{{ netbox_url }}"
        netbox_token:  "{{ netbox_token }}"
        data:
          termination_a_type: "dcim.interface"
          termination_a: 
            device: "Rxyy0001"
            name: "GE0/0/1"
          termination_b_type: "circuits.circuittermination"
          termination_b:
            circuit: "xyz-Primary"
            term_side: "A"
          status: "connected"
          type: "cat6"
        state: present
        validate_certs: false
```

### Expected Behavior

netbox_cable module is idempotent. No errors when executing the same task twice.

### Observed Behavior

Error as follows on the second run: 

```
fatal: [localhost]: FAILED! => {
    "changed": false,
[...]

PR #948, opened 2023-02-21
Ensure that nb_app is set before reference

sc68cal:


ryanmerolle:
relates to issue #946

PR #956, opened 2023-03-10
update device position to float

bk2zsto:
https://github.com/netbox-community/netbox/issues/51 breaking change from 3.3.0


sc68cal (changes requested):
You should create a new unit test that creates a position that is a float, instead of changing all the existing tests

PR #961, opened 2023-03-30
Add module netbox_journal_entry

rodvand:
Add module netbox_journal_entry to enable creation of journal entries on objects.

## Double Check


sc68cal (commented):
I'm a little iffy on the `assigned_object_id` parameter, since other modules like `netbox_ip_address`  and `netbox_cable` use a `device` parameter. Under the covers, it may end up resolving it to an ID and sending that to the REST API server, but this is the first module I've encountered where we are expecting an ID of int type. 

For example:

https://github.com/netbox-community/ansible_modules/blob/devel/plugins/modules/netbox_ip_address.py#L341

sc68cal:
If you add that example, demonstrating accessing the object's ID attribute to the docs I think that will resolve my concern. The original examples didn't have complex enough examples

PR #963, opened 2023-03-30
Fix 958: Location filter

### LINKED ISSUE #958: [Bug]: Nebox_location module updates existing locations instead of creating new ones in a different site

v3.11

### Ansible version

Running in AWX. Version is irrelevant to the case.

### NetBox version

3.3.7

### Python version

3.8

### Steps to Reproduce

When trying to create new location in a new site using netbox.netbox.netbox_location module updates existing locations with the same name, even though they are located in a different site, instead of creating new ones with their own separate site.

Locations are created like so:
```yaml
  - name: Creating locations
    netbox.netbox.netbox_location:
      data: "{{ location }}"
      state: present
    loop: "{{ latvia_locations }}"
    loop_control:
      loop_var: location
      label: "{{ location.site, location.name }}"
    tags: LV.locations
```
with looped locations defined in the following way:
```yaml
latvia_locations:
  - name: Marupe
    description: Main location
    site: Marupe
    tenant: xxx
    tags: "{{ latvia_defaults.tags }}"
  - name: Office Data Center
    description: Office building
    parent_location: Marupe
    site: Marupe
    tenant: xxx
    tags: "{{ latvia_defaults.tags }}"

```

### Expected Behavior

Existing locations with the same name but different sites should stay unchanged and previously undefined site/location pairs should be created properly instead.

### Observed Behavior

Because I already have a location named "Office Data Center" within a different site, e.g. "Riga", what actually happens is that the latter location in Riga gets updated (its site is effectively changed from Riga to Marupe) instead of a new one in Marupe being created.

rodvand:
# Fix #958
Add site as a possible filter when creating a location. You can now create a location named the same in two different sites.

## Double Check

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -435,7 +435,7 @@ ALLOWED_QUERY_PARAMS = {
     ),
     "l2vpn": set(["name"]),
     "lag": set(["name"]),
-    "location": set(["slug"]),
+    "location": set(["slug", "site"]),
     "module_type": set(["model"]),
     "manufacturer": set(["slug"]),
     "master": set(["name"]),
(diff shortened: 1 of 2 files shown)


sc68cal:
Can we add a set of tests for regression purposes

Issue #769, opened 2022-04-17
[Feature]: Add the module Module


rodvand:
### NetBox version

v3.2.1

### Feature type

New Module

### Proposed functionality

Add an Ansible module for the new NetBox feature Modules https://netbox.readthedocs.io/en/stable/core-functionality/modules/

### Use case

Be able to create/edit/delete and assign modules.

noziwatele:
Hi, wondering if any work has been done towards implementing a Module module?

PR #983, opened 2023-04-18
WIP

jgroom33:
## Related Issue

#769 

## New Behavior

* [ ] Add module Module
* [x] Add netbox_module_interface_template module

Self documenting

Add module_interface_template module
Add module module

## Double Check

### DIFF
--- a/plugins/module_utils/netbox_dcim.py
+++ b/plugins/module_utils/netbox_dcim.py
@@ -91,6 +91,7 @@ class NetboxDcimModule(NetboxModule):
         - inventory_item_roles
         - locations
         - manufacturers
+        - module_types
         - platforms
         - power_feeds
         - power_outlets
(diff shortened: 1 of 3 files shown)


jgroom33:
This is about as far as I'm going to take this PR.  I tried setting up the dev and test environment and it was a pain, so I stopped. I'll leave this PR open in case anyone wants to move it forward more. :) It's working for my use case.

One note: There seems to be a requirement for there to be a user defined key (slug, name, model, etc). The Module component doesn't have an obvious user defined key that could be used. So, I selected `asset_tag` even though it's not required property, it seemed to fit the purpose.

bar0n36:
We would love to see this functionality added, as it seems a bit of a gap. Currently for us to do this, we have to go and manually interact with the REST endpoints via the URI module, so effectively step out of the collection to do a couple of steps, and then step back in to continue (eg: Step 1 we add the device via the netbox_device module, then Step 2 we have to use the URI module to add module bays, Step 3 we have to do other stuff in the netbox collection again)

PR #998, opened 2023-05-10
Bump CI to use Ansible-Core 2.14

ryanmerolle:
Changed:
- Applied isort/ruff/black to all tests.
- Removed unneeded script logic in the python and bash scripts
- Removed NetBox 3.2 testing
- Added NetBox 3.5 testing (disabled in the action given it will fail currently)
- Updated ansbile-core/tests to 2.14
- Updated coverage as required by ansible-core 2.14
- Added python3.11 testing now that ansible-test 2.14 unblocked that
- Replaced jsondiff with deepdiff due to the out of order diffs breaking the integration tests


ryanmerolle:
Thanks to @FragmentedPacket doe helping me diagnose this issue so I could apply a fix.

tests/integration/netbox-deploy.py
```diff
@@ -33,7 +34,6 @@ def make_netbox_calls(endpoint, payload):
         created = endpoint.create(payload)
     except pynetbox.RequestError as e:
         print(e.error)
-        ERRORS = True
```
sc68cal, line 36: How come this was removed? No discussion in the commit message and this appears to alter behavior

sc68cal (changes requested):

tests/integration/netbox-deploy.py
sc68cal: Basically ERRORS now is always `False` - which means the logic the bottom that checks for `ERRORS` never executes even if there is an error.

ryanmerolle: I swore I did not delete it even though sourcery suggested it, thanks for the catch.

tests/integration/targets/v3.5/tasks/netbox_vm_interface.yml
```diff
@@ -54,14 +54,14 @@
...
-        - name: VoIP
-          site: Test Site
+      #untagged_vlan:
+      #  name: Wireless
+      #  site: Test Site
+      #tagged_vlans:
+      #  - name: Data
+      #    site: Test Site
+      #  - name: VoIP
+      #    site: Test Site
```
sc68cal, line 64: These are commented out, should they be removed? Or did the API change and that's why these are commented out?

sc68cal (changes requested):
Question about some commented out tests in `netbox_vm_interface`

PR #1023, opened 2023-06-12
Feat add possibility to use Bearer token

F-PMOR:
## Important information
this is my first contribution for an open source project. I'm not a full dev and have many withdraw with testing process, so I didn't do it with pytest, but I do manual test on my project. So I think some change will have to be done by expert, sorry and thank's for your understanding. 

## Related Issue

id : 1014

## New Behavior

This feature add the possibility to change the type of _token_ that has to be send to netbox or to a proxy behind netbox.
If the type of token is specified as **Bearer** (or **bearer**), it's send as a **Bearer** _token_ 

...

A check is done to verify the type of token configured in the ansible inventory file. If **Bearer** or **bearer** key is found, then **Bearer** is added before the key to send in Authorization headers. However, **Token** is the keyword send as usual.

...

Benefit : 
- **Bearer** token can now be used without effort.
- Ansible can now connect to Netbox that is protected by a proxy using for exemple Oauth2 authentification.
- **Token** key type still can be used without regression.

drawbacks 
- none

Changed has been done cause I use a security proxy behind Netbox that accept only **Bearer** Token.
For security reason, Token as to be renewed often, this is **not** managed by this feature change.
It can be done by updating the inventory file.
...

Some information like below should be added to documentation, but I don't know where ? 

_In the inventory file :_ 
Token : (string) should contain the token used to connect to netbox. If keyword **Bearer** is added at the beginning of the token, the plugins will use **Bearer**  token type instead of standard **Token** type.
exemple : 
token: thisismytoken # Token of type : Token
token: Bearer thisismytoken # token of type : Bearer
token: bearer thisismytoken # token of type : Bearer

...

Add **Bearer** token type.
...

## Double Check


sc68cal:
So, this `Bearer` token that you are adding to the header, is this an authentication type that is supported by NetBox, or a proxy specifically?

If this is something specific to your installation of netbox, and a proxy that you have set up to access your specific netbox that requires a bearer token, I don't know that we will want to accept this patch since it is specific to your environment

sc68cal (changes requested):
Appears to be a change that is to support a single use case of a specific install

F-PMOR:
> Appears to be a change that is to support a single use case of a specific install

Hello @sc68cal ,
Netbox support many authentification solution as describe here : https://demo.netbox.dev/static/docs/administration/authentication/
and here : https://python-social-auth.readthedocs.io/en/latest/
One of theme is Oauth2, and for Oauth2 works, we need to send a JWT (Json Web Token) to netbox via Authorization Bearer.
So it's not our solution but we use a solution supported by Django to make our Authentification more secure.
Regards, 
PM

rodvand:
I think this is a valid PR. @F-PMOR Could you just run black to make the formatting ok? Just run `./hacking/black.sh` while having the virtual environment enabled.

F-PMOR:
> I think this is a valid PR. @F-PMOR Could you just run black to make the formatting ok? Just run `./hacking/black.sh` while having the virtual environment enabled.

Ok, Thanks @rodvand , I will do it today, I'll send you the result or my miss knowledge of the process...  Thank you @sc68cal, for the exchange that permit me to update the comments to argue the RF, Regards

F-PMOR:
Hello @rodvand, check pass successfully, does I have something else do to ?
Thanks,
Philippe

sc68cal:
This needs documentation added so that people understand how to utilize

F-PMOR:
> This needs documentation added so that people understand how to utilize

Thanks @sc68cal , where can I put documentation ? It's my first PR and I'm a little bit lost. I put some information to add to the documentation in my PR. Is it enought ? 
Regards,
Philippe

sc68cal:
No, you need to at the very least add it to the documentation for the `token` parameter at the top of the file you have changed, line 83 specifically.

plugins/inventory/nb_inventory.py
```diff
@@ -2015,7 +2029,7 @@ def parse(self, inventory, loader, path, cache=True):
         self.ca_path = self.get_option("ca_path")
         if token:
             # add possibility to use Bearer token (JWT)
-            if "bearer" in token.lower():
+            if "bearer" in token.lower():pass
```
sc68cal:
```suggestion
            if "bearer" in token.lower():
```

sc68cal (dismissed):

F-PMOR:
> Honestly I think this needs some work. I think I would rather see `token` changed to be a structure with a `type` and `value` attribute, with the `type`'s default being `token` for backwards compatibility
> 
> Basically I don't like the string manipulation that you are doing, and would prefer it to be something like:
> 
> ```python
> self.headers.update({"Authorization": f"{{token.type}} {{token.value}}"})
> ```
> 
> So that we could support the following in configuration:
> 
> ```
> token:
>   type: Bearer
>   value: thisismytoken
> ```
> 
> and
> 
> ```
> token:
>    type: Token
>    value: thisismytoken
> ```

Yes, I agree, it's better, but I'm not able to do that ! I ever don't know what's happen and how works the CI linter for Documentation that doesn't pass.
Can you change the code on your side to have something better and taht's work ? I really need help !
Thanks.

sc68cal:
Yep that's fair, I will see if I can add commits to your PR, and then do some research to try and accomplish what I was thinking

F-PMOR:
thanks a lot  @sc68cal  and @rodvand for this Pull Request.
Regards,
Philippe

PR #1028, opened 2023-06-23
Update pull_request_template.md

sc68cal:
Adds a note about having a unit test


sc68cal:
Might need to work on the wording of this since there are cases where existing unit tests could exercise new code paths without modification.

Issue #370, opened 2020-10-29
nb_lookup: Can't query IP-Address by id


idle-hae:
##### ISSUE TYPE
 - Bug Report

###### Ansible:
2.9.13

###### Netbox:
2.9.4

###### Collection:
netbox collection version 1.1.0

##### SUMMARY
I need to query netbox for a specific ip address. I would like to filter this with the id of the ip-address.
This results in an error.

##### STEPS TO REPRODUCE

```yaml
- name: "Get ip-address by id"
  debug:
    msg:  "{{ query('netbox.netbox.nb_lookup', 'ip-addresses',
                    api_endpoint=netbox_url,
                    api_filter='id=6293',
                    token=netbox_token) }}"
```

##### EXPECTED RESULTS
I expect to get a dict of the queried ip address.

##### ACTUAL RESULTS
```
An unhandled exception occurred while running the lookup plugin 'netbox.netbox.nb_lookup'. Error was a <class 'ValueError'>, original message: A reserved ('id', 'pk', 'limit', 'offset') kwarg was passed. Please remove it try again.
```

FragmentedPacket:
This appears to be an implementation detail with using `pynetbox` and is generated from that library.

It appears we should be able to query by `id`, but not sure the why on not searching by it.

I would open an issue with `pynetbox` and see the reasoning and see if there is a possible workaround.

idle-hae:
Im sorry to open the issue again.

Is it possible that I'm just using the lookup module wrong for what I'm trying to achieve? Maybe there is just an argument, other than api_filter, that I need to use.

I noticed, that one doesn't need to search for a specific address by id, but can define it as part of the path. (e.g. http://netbox/api/ipam/ip-addresses/6293/)

What the api filter does is calling http://netbox/api/ipam/ip-addresses/?id=6293, which leads to a list with one entry.
Is there a way in the lookup module to not search for the address but to query by id? This should return just the specific object and not a list.

According to the documentation I found pynetbox should support this.
https://pynetbox.readthedocs.io/en/latest/endpoint.html
Try to search for "Referencing with an id."

Is there a way to call pynetbox this way with the lookup module?

FragmentedPacket:
I've submitted a PR for this. Are you able to test this?

idle-hae:
Thanks for reconsidering the issue!

Your PR looks like its working.
I tried to add some more parameters after `id` and it seems to handle that gracefully (pynetbox seems to just ignore them).

[...]

PR #1042, opened 2023-07-28
Fix bug in nb_lookup if api_filter contains multiple `id`

linkdd:
## Related Issue

 - #370
 - #376

## New Behavior

In the `nb_lookup` plugin, when supplying multiple `id` fields to the `api_filter` parameter, it will return a list of items.

In the `nb_lookup` plugin, if we provide an `api_filter` containing the `id` field, it will use `.get()` instead of `.filter()`, thus returning the item instead of a list. See PR #376 for more information.

This is fine if you supply only one id. But let's say you want to get a list of items by ids?

For example: `id=1 id=5`

I would expect to get a list of 2 items. But at the moment, it will return the first item only.

This change adds a simple check to see if we're supplying multiple ids.

This fixes a bug where supplying multiple ids in the `api_filter` ends up returning only one item.

In fact, I'd even expect that supplying a single id would return a list of one item, since we're querying a collection. So maybe a `many=True|False` parameter would be better?

This bugfix as it is is backward-compatible.

## Double Check

### DIFF
--- a/plugins/lookup/nb_lookup.py
+++ b/plugins/lookup/nb_lookup.py
@@ -439,7 +439,7 @@ class LookupModule(LookupBase):
             if netbox_api_filter:
                 filter = build_filters(netbox_api_filter)
 
-                if "id" in filter:
+                if "id" in filter and len(filter["id"]) == 1:
                     Display().vvvv(
                         "Filter is: %s and includes id, will use .get instead of .filter"
                         % (filter)


sc68cal:
At the very least, this needs a test that exercises sending two IDs as a filter and verifying the desired behavior

sc68cal (changes requested):
missing new integration test in https://github.com/netbox-community/ansible_modules/blob/devel/tests/integration/targets/v3.5/tasks/netbox_lookup.yml

PR #1053, opened 2023-08-16
Closes #1052: Schema caching is broken

### LINKED ISSUE #1052: [Bug]: Schema caching is broken

v3.13.0

### Ansible version

```bash
ansible [core 2.15.1]
  config file = /Users/decoupca/code/ansible/ansible-dev.cfg
  configured module search path = ['/Users/decoupca/code/ansible/library']
  ansible python module location = /Users/decoupca/code/ansible/venv/lib/python3.11/site-packages/ansible
  ansible collection location = /Users/decoupca/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/decoupca/code/ansible/venv/bin/ansible
  python version = 3.11.4 (main, Jun 20 2023, 17:23:00) [Clang 14.0.3 (clang-1403.0.22.14.1)] (/Users/decoupca/code/ansible/venv/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True
```

### NetBox version

v3.5.6

### Python version

3.11

### Steps to Reproduce

Use `nb_inventory` plugin to run a playbook more than once. The play doesn't matter, only that `nb_inventory` is used as an inventory source.

### Expected Behavior

On the first run, `nb_inventory` fetches the API schema at `api/schema/?format=json` and caches it to `~/.ansible/tmp/netbox_api_dump.json`. Future runs using the same version of NetBox should use this cached copy. This is important for performance because the schema file is in many cases very large (>6MB in my testing), so retrieving it every run degrades performance significantly.

### Observed Behavior

`nb_inventory` fetches the API schema on every run, regardless of the fact that A) the version of NetBox hasn't changed, and B) there is a cached copy in `~/.ansible/tmp`.

decoupca:
## Related Issue

#1052 

## New Behavior

Fix schema caching to align with intended behavior: download only once per major NetBox version.

...

Schema caching is broken, i.e., re-fetches the (extremely large) schema on every playbook run.

...

The schema file is >6MB, so re-downloading it on every run is time-consuming, even on high performance networks.

...

No changes, as this is fixing intended behavior.

...

Fix API schema caching.

...

## Double Check

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -1505,7 +1505,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         except Exception:
             openapi = {}
 
-        cached_api_version = ".".join(openapi.get("info", {}).get("version").split(".")[:2])
+        cached_api_version = ".".join(
+            openapi.get("info", {}).get("version").split(".")[:2]
+        )
 
         if netbox_api_version != cached_api_version:
             if version.parse(netbox_api_version) >= version.parse("3.5.0"):
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -1505,7 +1505,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         except Exception:
             openapi = {}
 
-        cached_api_version = openapi.get("info", {}).get("version")
+        cached_api_version = ".".join(openapi.get("info", {}).get("version").split(".")[:2])
 
         if netbox_api_version != cached_api_version:
             if version.parse(netbox_api_version) >= version.parse("3.5.0"):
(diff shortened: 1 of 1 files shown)


plugins/inventory/nb_inventory.py
```diff
@@ -1505,7 +1505,8 @@ def fetch_api_docs(self):
         except Exception:
             openapi = {}
 
-        cached_api_version = openapi.get("info", {}).get("version")
+        if cached_api_version := openapi.get("info", {}).get("version"):
+            cached_api_version = ".".join(cached_api_version.split(".")[:2])
```
sc68cal, line 1509: I would prefer to avoid using the `:=` operator. This would be the first usage in our entire repo, so would you be willing to rewrite it, to not use `:=`? Thanks

sc68cal (changes requested):
style concern

plugins/inventory/nb_inventory.py
decoupca: You bet, done.

PR #1062, opened 2023-09-02
Fix device_role to role breaking change in netbox_device module

shaul75:
Updated the Ansible NetBox module to accommodate recent breaking changes in the NetBox API where `device_role` has been renamed to `role`. Also made corresponding adjustments in the pynetbox library.

## New Behavior

...

...

...

...

...

## Double Check

* [ ] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netbox-community/ansible_modules/blob/devel/CONTRIBUTING.md).
* [ ] I have explained my PR according to the information in the comments or in a linked issue.
* [ ] My PR targets the `devel` branch.

### DIFF
--- a/plugins/modules/netbox_device.py
+++ b/plugins/modules/netbox_device.py
@@ -40,7 +40,7 @@ options:
           - Required if I(state=present) and the device does not exist yet
         required: false
         type: raw
-      device_role:
+      role:
         description:
           - Required if I(state=present) and the device does not exist yet
         required: false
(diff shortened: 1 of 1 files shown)


rodvand:
We can't just stop supporting 3.5 and 3.4, so the device_role and role will have to live side by side for a while. We should also note about deprecation on the device_role field.

FragmentedPacket:
Hi @shaul75,

Thanks for your PR and taking the initiative to submit a fix! As @rodvand stated, we still need to support previous versions so logic must be added  to make this change dependent on the version of Netbox.

I created https://github.com/netbox-community/ansible_modules/pull/1066 with the fix for backwards compatibility if you want to check it out.

If you'd like, I can submit a PR into your fork based off of your PR to ensure you get credit for the fix as well. This would obviously delay the fix dependent on our schedules, but figured I'd at least ask.

Issue #1032, opened 2023-07-05
[Bug]: Inventory Parsing Fails in Current Redhat Python Build


HeroesLament:
### Ansible NetBox Collection version

v3.13.0

### Ansible version

```bash
ansible-inventory [core 2.15.1]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.11/site-packages/ansible
  ansible collection location = /runner/requirements_collections:/root/.ansible/collections:/usr/share/ansible/collections:/usr/share/automation-controller/collections
  executable location = /usr/local/bin/ansible-inventory
  python version = 3.11.2 (main, May 24 2023, 00:00:00) [GCC 11.3.1 20221121 (Red Hat 11.3.1-4)] (/usr/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True
```

### NetBox version

v3.4.7

### Python version

3.11

### Steps to Reproduce

ansible-inventory -i netbox.yml --list

### Expected Behavior

```console
{
    "_meta": {
        "hostvars": {
            "rtr-lhr-01": {
                "cluster": "London 1",
                "cluster_type": "vultr",
                "custom_fields": {},
                "device_roles": [
                    "router"
                ],
                "disk": 60,
                "is_virtual": true,
                "local_context_data": [
                    null
                ],
                "locations": [],
                "memory": 2048,
                "regions": [],
                "services": [],
                "site_groups": [],
                "sites": [
                    "vultr-lhr-01"
                ],
                "status": {
                    "label": "Active",
                    "value": "active"
                },
                "tags": [
                    "production"
                ],
                "vcpus": 2.0
            }
        }
    },
    "all": {
        "children": [
            "ungrouped",
            "tags_production"
        ]
    },
    "tags_production": {
        "hosts": [
            "rtr-lhr-01"
        ]
    }
}
```

### Observed Behavior

``` console
setting up inventory plugins
Loading collection ansible.builtin from 
host_list declined parsing /runner/project/tools/ansible-awx/netbox.yml as it did not pass its verify_file() method
script declined parsing /runner/project/tools/ansible-awx/netbox.yml as it did not pass its verify_file() method
Loading collection netbox.netbox from /usr/share/ansible/collections/ansible_collections/netbox/netbox
[...]

PR #1077, opened 2023-09-30
Raise exception instead of setting invalid netbox api version

sc68cal:
This method would previously set netbox_api_version to 0, which is not a valid version that can be parsed. 

This does not fix the issue, but raises this exception with a better message to help the user diagnose

See #1032


sc68cal:
I might change this to raise an AnsibleError or something similar so that it emits a better log message

PR #1066, opened 2023-09-06
Fix device_role -> role and add testing for Netbox 3.6

FragmentedPacket:
## Related Issue
`device_role` was renamed to `role` within the Netbox API for v3.6+
Filter for IP Addresses now require `virtual_chassis_member_id` to fetch VC member interfaces and `device_id` no longer returns the VC member interfaces.

Fixes #1061

## New Behavior

Compatibility with Netbox APIs but no user changes need to occur.

## Proposed Release Note Entry

Fix `device_role` being renamed to `role` in Netbox 3.6+ API
Change query params for searching interfaces when we need to find the VC member interfaces.

Issue #1061, opened 2023-09-02
[Bug]: Breaking changes in 3.6.0, device_role -> role


shaul75:
PR: https://github.com/netbox-community/ansible_modules/pull/1062

### Ansible NetBox Collection version

v3.14.0

### Ansible version

https://github.com/netbox-community/netbox/releases/tag/v3.6.0

> The device_role field on the Device model has been renamed to role. The device_role field has been temporarily retained on the REST API serializer for devices for backward compatibility, but is read-only.

### NetBox version

v3.6.0

### Python version

3.10

### Steps to Reproduce

```
      netbox.netbox.netbox_device:
        data:
          name: "{{ ansible_facts['product_name'] }}"
          device_type: "{{ ansible_facts['product_name'] }}"
          device_role: 1
          site: 1
        state: present

```

### Expected Behavior

Device is created/updated

### Observed Behavior

```
{"role":["This field is required."]}
fatal: [10.0.20.10 -> localhost]: FAILED! => {
    "changed": false
}
```

sthierolf:
Can confirm bug also with NetBox 3.6.1.
By manually adding the device (this means: device already exists) no error is thrown.

Should be in relation to 3.6.0 (and higher) releases, quote:
_The device_role field on the Device model has been renamed to role. The device_role field has been temporarily retained on the REST API serializer for devices for backward compatibility, but is read-only._

https://github.com/netbox-community/netbox/releases/tag/v3.6.0

gtherond:
I do confirm, the issue comes from the following upstream API breaking change: https://github.com/netbox-community/netbox/issues/6391

They did put device_role field as a RO one while the RW field is now named role.
A simple quick and dirty fix would be to change the device_role mapping to role instead of device_role api field.
in here: https://github.com/netbox-community/ansible_modules/blob/4211b713c07d45c69f14c94dc1894cbb813589a5/plugins/module_utils/netbox_utils.py#L315

PR #1080, opened 2023-10-03
fix: use 'role' instead of 'device_role' with netbox >= 3.6.0 in netbox_device module

floatingstatic:
## Related Issue

I have the same issue as what this PR was attempting to address:

https://github.com/netbox-community/ansible_modules/pull/1062

## New Behavior

When talking to Netbox >= 3.6.0 replace `device_role` parameters sent to the netbox API with `role`. This is needed due to breaking changes introduced in 3.6.0 where `device_role` was renamed to `role`. Ref: https://github.com/netbox-community/netbox/releases/tag/v3.6.0

This not an attempt to explicitly add a new `role` parameter or deprecate `device_role` as a parameter. Instead this is a short-term fix for those of us who are using netbox >= 3.6 and want to continue to use this ansible collection with no changes to parameters passed. So basically we continue to pass `device_role` as a parameter even though this is technically no longer valid for for NB >= 3.6. This ensures backwards compatibility and no change in behavior for those on newer releases.

...

Benefits: allows the `netbox_device` module to work with NB 3.6 and later

Drawbacks: doesn't solve the problem of how to address this breaking change with a new parameter (`role`) to the ansible module directly while maintaining backwards comat. I defer to the collection authors for consensus on this (also happy to help if you can provide some guidance on how you would like this to work).

N/A for now

Make 'device_role' parameter compatible with Netbox >= 3.6.0 in netbox_device module

## Double Check

### DIFF
--- a/plugins/module_utils/netbox_dcim.py
+++ b/plugins/module_utils/netbox_dcim.py
@@ -210,6 +210,13 @@ class NetboxDcimModule(NetboxModule):
                 nb_endpoint, object_query_params, name
             )
 
+        if (
+            self.endpoint == "devices"
+            and data.get("device_role")
+            and Version(self.full_version) >= Version("3.6.0")
+        ):
+            data["role"] = data.pop("device_role")
+
         # This is logic to handle interfaces on a VC
         if self.endpoint == "interfaces" and self.nb_object:
             child = self.nb.dcim.devices.get(self.nb_object.device.id)


floatingstatic:
Oops apologies I missed https://github.com/netbox-community/ansible_modules/pull/1066 which is much more comprehensive. Closing in favor of that since it is already merged anyway!

Related, when will the new version of this module collection be updated so we can start taking advantage of those changes? Thanks!

lasag-2511:
"Unsupported parameters for (netbox.netbox.netbox_device) module: data.role. i still have issues with it

floatingstatic:
@lasag-2511 This was fixed in #1066 . Prior to the release `3.15.0` I ended up building from develop as `3.14.0` would not have had the fix. Something like:

```
git clone https://github.com/netbox-community/ansible_modules
ansible-galaxy collection build .
ansible-galaxy collection install netbox-netbox*.tar.gz
```

At this point you can just upgrade to version `3.15.0` of the collection and it works fine (at least for me).

lasag-2511:
> @lasag-2511 This was fixed in #1066 . Prior to the release `3.15.0` I ended up building from develop as `3.14.0` would not have had the fix. Something like:
> 
> ```
> git clone https://github.com/netbox-community/ansible_modules
> ansible-galaxy collection build .
> ansible-galaxy collection install netbox-netbox*.tar.gz
> ```
> 
> At this point you can just upgrade to version `3.15.0` of the collection and it works fine (at least for me).

hi we are not using ansible-galaxy , what to do in that case?

lmas-chwy:
I installed version 3.16.0 and still have the same error

floatingstatic:
@lmas-chwy why not comment on #1061 since you are commenting on a PR I opened that was never merged and addressed and in my opinion fixed in in #1066. A PR that was never merged is not an issue that will get any visibility by maintainers. Thanks!

PR #1114, opened 2023-11-08
Decrypt ansible-vault strings in nb_inventory

jetomit:
This allows storing secrets in config context data.

## Related Issue

#551

## New Behavior

This allows storing ansible-vault secrets in a NetBox config context, like this:

    {
        "plain_var": "value",
        "secret_var": {
            "__ansible_vault": "$ANSIBLE_VAULT;1.1;AES256\n123456…"
        }
    }

Currently encrypted values are returned as is, and cannot be used directly in Ansible playbooks or templates without the workaround described in #551.

The change should have no effect on existing installations; the workaround in #551 will still work but do nothing, as the values are now decrpyted earlier. Other possible workarounds (e.g. decrypting values in a playbook) might break if they expect an ansible-vault string.

Replacing `json.loads` with `self.loader.load` makes `nb_inventory` work more like [the builtin `script` plugin](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/inventory/script.py#L106-L115), so I believe it is a better solution. If additional functionality is added to the `DataLoader` class in Ansible, it should work with `nb_inventory` with no changes to the plugin.

No changes. Please let me know if I should add a note somewhere.

Support ansible-vault encrypted values in NetBox config contexts

## Double Check

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -453,7 +453,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 )
 
             try:
-                results = json.loads(raw_data)
+                results = self.loader.load(raw_data, json_only=True)
             except ValueError:
                 raise AnsibleError("Incorrect JSON payload: %s" % raw_data)
 


sc68cal:
Can you add a test that exercises this? Creat a device and a config context with an encrypted string, and successfully decrypt?

Issue #1135, opened 2023-12-22
[Feature]: NetBox 3.7 Custom Field UI visibility changes


bluikko:
### NetBox version

v3.7-beta1

### Feature type

Change to existing Module

### Proposed functionality

NetBox 3.7 changes the single `ui_visibility` field to 2 separate fields `ui_visible` and `ui_editable`.

Support the changes in a backwards-compatible manner by just adding the new fields and leaving the old field in place too.
PR in progress.

### Use case

Support features in the upcoming NetBox 3.7.

### External dependencies

None.

PR #1136, opened 2023-12-22
Support CustomField UI visibility in 3.8

bluikko:
## Related Issue

#1135

## New Behavior

Add to CustomField the 2 new UI visibility parameters.

Add ui_editable and ui_visible parameters that are new in NetBox 3.8.

Just add the new parameters and keep the old `ui_visibility` for backwards
compatibility with pre-3.7.

Added the 2 new parameters to module documentation.

Support CustomField UI visibility in NetBox 3.8.

## Double Check

### DIFF
--- a/plugins/modules/netbox_custom_field.py
+++ b/plugins/modules/netbox_custom_field.py
@@ -106,9 +106,30 @@ options:
         required: false
         type: str      
         version_added: "3.10.0"
+      ui_editable:
+         description:
+           - Whether the custom field is editable in the UI
+         required: false
+         choices: 
+           - yes
+           - no
+           - hidden
+         type: str      
+         version_added: "3.17.0"
+      ui_visible:
+         description:
+           - Whether the custom field is displayed in the UI
+         required: false
+         choices: 
+           - always
+           - if-set
+           - hidden
+         type: str      
+         version_added: "3.17.0"
       ui_visibility:
          description:
            - The UI visibility of the custom field
+           - Replaced by C(ui_visible) and C(ui_editabl)e in NetBox 3.7
          required: false
          choices: 
            - read-write
@@ -245,6 +266,24 @@ def main():
                     weight=dict(required=False, type="int"),
                     search_weight=dict(required=False, type="int"),
                     group_name=dict(required=False, type="str"),
+                    ui_editable=dict(
+                        required=False,
+                        choices=[
+                            "yes",
+                            "no",
+                            "hidden",
+                        ],
+                        type="str",
+                    ),
+                    ui_visible=dict(
+                        required=False,
+                        choices=[
+                            "always",
+                            "if-set",
+                            "hidden",
+                        ],
+                        type="str",
+                    ),
                     ui_visibility=dict(
                         required=False,
                         choices=[
--- a/plugins/modules/netbox_custom_field.py
+++ b/plugins/modules/netbox_custom_field.py
@@ -111,8 +111,8 @@ options:
            - Whether the custom field is editable in the UI
          required: false
          choices: 
-           - yes
-           - no
+           - "yes"
+           - "no"
            - hidden
          type: str      
          version_added: "3.17.0"
--- a/plugins/modules/netbox_custom_field.py
+++ b/plugins/modules/netbox_custom_field.py
@@ -129,7 +129,7 @@ options:
       ui_visibility:
          description:
            - The UI visibility of the custom field
-           - Replaced by C(ui_visible) and C(ui_editabl)e in NetBox 3.7
+           - Replaced by O(ui_visible) and O(ui_editable) in NetBox 3.7
          required: false
          choices: 
            - read-write
--- a/docs/plugins/netbox_custom_field_module.rst
+++ b/docs/plugins/netbox_custom_field_module.rst
@@ -619,6 +619,98 @@ Parameters
   * - .. raw:: html
 
         <div class="ansible-option-indent"></div><div class="ansible-option-cell">
+        <div class="ansibleOptionAnchor" id="parameter-data/ui_editable"></div>
+
+      .. _ansible_collections.netbox.netbox.netbox_custom_field_module__parameter-data/ui_editable:
+
+      .. rst-class:: ansible-option-title
+
+      **ui_editable**
+
+      .. raw:: html
+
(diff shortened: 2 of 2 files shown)


bluikko:
Tests are missing at this point. I see that they will need to be changed also.

bluikko:
Not sure what to do with the CI sanity test failure. I get that Ansible "helpfully" treats yes/true & no/false as the same.  
But these are not boolean values they are strings, and in the module doc block listed accordingly as `string`.

sc68cal:
YAML converts `yes` and `no` to boolean fields, you cannot use those values as strings. Consider changing it to a boolean and just supporting true/false and not have the `hidden` value

bluikko:
> Consider changing it to a boolean and just supporting true/false and not have the `hidden` value

It would be very disappointing to not have full feature parity especially since I would definitely want to use the `hidden` feature.
I believe a workaround would be to split this into 2 different boolean parameters:

1. `ui_editable`.
2. `ui_editable_hidden`.

I have opened a probably futile discussion item in netbox-community/netbox, let's see if there is any feedback before going down the route of 2 boolean parameters.

plugins/modules/netbox_custom_field.py
```diff
@@ -106,9 +106,30 @@
         required: false
         type: str      
         version_added: "3.10.0"
+      ui_editable:
+         description:
+           - Whether the custom field is editable in the UI
+         required: false
+         choices: 
+           - "yes"
+           - "no"
```
sc68cal, line 115: These values are going to cause a lot of confusion, people are going to just use `yes` and `no` without strings and YAML will convert them to booleans, and then the module will throw an error and they'll come here asking why it doesn't work.

bluikko: **Exactly** and that is what my comments above are about.  
I 100% agree.

I need feedback on what is the preferred solution to work around this. I do not think forcing users to quote values to *just one* module parameter is reasonable, at all.

Two other possible options:

1. Present 2 different boolean parameters instead of presenting `ui_editable` as a string value: `ui_editable` and `ui_editable_hidden`. Based on the combined value of these 2 parameters decide if give `yes` `no` or `hidden` to NetBox API.
2. Use some other values than `yes` and `no` as values `ui_editable`: module parameters takes other values for example `read-write` and calls NetBox API with `yes`; value `read-only` means `no` and `hidden` stays the same.

Open to other suggestions or feedback on these 2. I had an open discussion item at the netbox project and suggestions to not use `yes` and `no` in the NetBox API were ignored.

Both will make the module API diverge from NetBox API which will be sad.

rodvand: I think Option 1 of the other options is the best one. If we're diverging from the NetBox API we may as well do it the best way we can. Thanks for your work on this @bluikko

sc68cal: I also think option 1 is the best solution. I'm not crazy about this tri-boolean that the netbox API is exposing.

bluikko: Will rework this to 2x boolean then.

Jacob-gr: In looking to address the problem of `ui_visible` and `ui_editable`, I came across this old pull request.

I understand the concern of yes and no values being treated as booleans when not wrapped in quotes. However, when I was testing this out, it appears that as long as the `argument_spec` dict is declaring `ui_editable` and `ui_visible` as a `type=str` in `netbox_custom_field.py`, it will convert the yes and no values into strings automatically.  

To examine it further, I logged module output and observed the following:

When using `ui_editable: no` (declared as a bool) in the Ansible task, the `module.params` in `netbox_custom_field.py` ended up as a string of 'no':

```log
2025-01-02 14:55:10,774 - __main__ (netbox_custom_field.py) - DEBUG - Module params: {'netbox_url': 'https://netbox.example.io', 'netbox_token': 'SomeToken', 'data': {'description': 'Security Onion Hostname', 'filter_logic': 'exact', 'group_name': 'A Group', 'label': 'Hostname', 'name': 'somehostname', 'object_types': ['virtualization.virtualmachine'], 'required': True, 'type': 'text', 'ui_visible': 'if-set', 'ui_editable': 'no', 'validation_regex': '^[a-z0-9\\-]+$', 'weight': 5, 'content_types': None, 'object_type': None, 'default': None, 'search_weight': None, 'ui_visibility': None, 'validation_minimum': None, 'validation_maximum': None, 'choices': None}, 'state': 'present', 'validate_certs': True, 'query_params': None, 'cert': None}
```

When converting the the type to a `ui_editable` type to a bool and providing `choices=[False, True]`, only then was `ui_editable: no` treated as a bool and appeared as `False` in the module params.

I'm new to using Ansible, so I'm not sure if this was a change in the library or if my testing is flawed, but it appears there shouldn't be an issue even if users use `no` or `yes` without quotes. In light of this, I'd like to be able to just use yes, no, and hidden as it is documented in the API. Similar with ui_visibile: I'd prefer just to have the API values available.

Would this information contribute to this pull request being approved?

Versions and Misc.:
- Netbox version --> NetBox Community v4.1.10 (2024-12-23) 
- ansible version --> ansible [core 2.18.1]
- netbox.netbox version --> 3.20.0
- Had to update the `required_if` block in `netbox_custom_field.py` to be `("state", "present", ["object_types", "name"])` instead of `("state", "present", ["content_types", "name"])`. Maybe a change due to more recent netbox version? content_types wasn't useable for me and produced an error.

dhoffend:
I've tested the pull request (download 3.21.0 + manually applied the 4 patches). It works for me and I can finally use the module again, Cause the newer netbox version will reject updates with the `ui_visibility` attribute included

    ui_visibility does not exist on existing object. Check to make sure valid field.

Without this pull request beeing applied the module is partly broken

sidenotes:
* The fields `unique` and `is_cloneable` are also missing. But that's a different story

LaZyDK:
Can we soon have this merged so that we can use `ui_visible: if-set`?

plugins/modules/netbox_custom_field.py
Andres1357: Any shot of this pull request being merged?

dhoffend:
@arthanson could you maybe take a look into this PR? This ansible module is broken since netbox 3.8

Issue #852, opened 2022-10-21
[Feature]: Add vm assigned device to inventory vars


fprina:
### NetBox version

v3.3.5

### Feature type

Change to existing Module

### Proposed functionality

In the inventory variables related to a virtual machine should be present the assigned device

According to the web interface restrictions I suppoe the right var name is `cluster_device`

For esample [vm1](https://demo.netbox.dev/virtualization/virtual-machines/361/) in the demo site is assigned to a specific device so `ansible-inventory --host vm1` should return:

{
    "cluster": "DO-AMS3",
    "**cluster_device": "NCblade[0-3]"**
    "cluster_group": "emea",
    "cluster_type": "digitalocean",
    "custom_fields": {},
    "device_roles": [
        "application-server"
    ],
    "is_virtual": true,
    "local_context_data": [
        null
    ],
    "locations": [],
    "platforms": [
        "ubuntu-linux-20-04"
    ],
    "regions": [],
    "services": [],
    "site_groups": [],
    "status": {
        "label": "Active",
        "value": "active"
    },
    "tags": []
}

### Use case

use the assigned device directly in the ansible playbooks

joek-office:
Nice enhancement request, 
is anyone out there who is working on this?

PR #1169, opened 2024-02-13
add support to fetch cluster device information from netbox.

joek-office:
## Related Issue

#852 

## New Behavior

in ansible netbox inventory will exist a new hostvar with name cluster_device if corresponding information is set in netbox.

...

currently the information is missing in the netbox inventory

...

Benefit: More Information can be used from netbox inventory
drawback: i see no one.
backwards-compatible: i think yes, but can't test how far in the past it gets.
i need this information to configure high availability and replication tasks in proxmox over api.

...

no changes needed in my opinion

...

add support to fetch cluster device information for virtual machine resources from netbox.

...

## Double Check

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -537,6 +537,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             "custom_fields": self.extract_custom_fields,
             "region": self.extract_regions,
             "cluster": self.extract_cluster,
+            "cluster_device": self.extract_cluster_device,
             "cluster_group": self.extract_cluster_group,
             "cluster_type": self.extract_cluster_type,
             "is_virtual": self.extract_is_virtual,
@@ -940,6 +941,13 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         try:
             # cluster does not have a slug
             return host["cluster"]["name"]
+        except Exception:
+            return
+
+    def extract_cluster_device(self, host):
+        try:
+            # cluster device does not have a slug
+            return host["device"]["name"]
         except Exception:
             return
 


plugins/inventory/nb_inventory.py
```diff
@@ -943,6 +944,13 @@ def extract_cluster(self, host):
         except Exception:
             return
 
+    def extract_cluster_device(self, host):
+        try:
+            # cluster device does not have a slug
+            return host["device"]["name"]
+        except Exception:
```
sc68cal, line 951: Is there a more specific exception we could use like KeyError?

sc68cal (changes requested):

plugins/inventory/nb_inventory.py
sc68cal: Could this also use a more common pattern of checking if the key exists before accessing, instead of a try/except block?

joek-office: I don't know if i have understand it right, but have developed a solution without try/catch and with returning a suitable value in every case.
New version see below

plugins/inventory/nb_inventory.py
```diff
@@ -943,6 +944,12 @@ def extract_cluster(self, host):
         except Exception:
             return
 
+    def extract_cluster_device(self, host):
+        # cluster device does not have a slug
+        if host.get("device", None) is None:
+            return ""
```
sc68cal, line 950: Are you certain this is the correct thing to return? This doesn't seem like a good idea to me

plugins/inventory/nb_inventory.py
```diff
@@ -943,6 +944,12 @@ def extract_cluster(self, host):
...
+        return host.get("device", {"name": ""}).get("name", "")
```
sc68cal, line 951: This looks awful similar to the above return when there is no device key in the dictionary. 

Overall I think this needs to be improved

joek-office: What is your opinion should we return, if there is no value assigned. Should it be null/None?
Also, is there a handy way to determine the object type of the current host. Then i can only return the cluster_device variable, when the object is an virtual machine.

sc68cal: > What is your opinion should we return, if there is no value assigned. Should it be null/None?

I would follow the pattern set by the other functions in this code. Specifically it appears that most would `return` (which is short hand for returning None) although some return an empty list where it makes sense to.

sc68cal (changes requested):
Remove the changes to gitignore and remove the install.sh script

joek-office:
> @joek-office I'll try and take a look next week, I'm a bit swamped at the moment. I agree that's an unusual error.
Any news so far?

sc68cal:
It's taking me a bit of time to transfer to a new machine to do development work, bear with me

joek-office:
@sc68cal 
the question comes more from my side to don't forget the PR and to track progress. Please don't feel in any hurry from my request.

sc68cal:
I'm going to re-run the failing jobs and see if it was just a transient error

joek-office:
Looks like a persistent error. Any ideas to solve the errors?

Issue #1185, opened 2024-03-07
[Bug]: Lookup plugin not supporting custom-fields-choice-sets


MrDiba:
### Ansible NetBox Collection version

v3.17.0

### Ansible version

```bash
ansible [core 2.13.2]
  config file = /home/ansible/ansible/ansible.cfg
  configured module search path = ['/home/ansible/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ansible/.local/lib/python3.9/site-packages/ansible
  ansible collection location = /home/ansible/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/ansible/.local/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
  jinja version = 3.1.2
  libyaml = True
```

### NetBox version

v3.7.3

### Python version

3.9

### Steps to Reproduce

Create a Custom Field Choice Sets called set1

Run the following playbook:
---
- name: "Lookup choice set"
  hosts: localhost
  tasks:

    - name: debug
      debug:
        msg: "{{ (query('netbox.netbox.nb_lookup', 'custom-field-choice-sets', api_filter='name=set1', api_endpoint=netbox_url, token=netbox_token ))[0] }}"

### Expected Behavior

I expect an list with information about the choice set, including the choices.

### Observed Behavior

```
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'netbox.netbox.nb_lookup'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Unrecognised term custom-field-choice-sets. Check documentation. Unrecognised term custom-field-choice-sets. Check documentation"}
```

MrDiba:
Related #1069

MrDiba:
Pull request is merged to the development branch. Closing this issue.

PR #1186, opened 2024-03-07
Added custom-field-choice-sets to nb_lookup.py

MrDiba:
## Related Issue

#1185 

## New Behavior

Since NetBox v3.6.0 there is an breaking change where the custom-fields and choices are split.

Currently there is no way to access the custom field choices when using NetBox v3.6.0 or higher

_Why do you think this project and the community will benefit from your proposed change?_

Starting from NetBox v3.6.0 it is not possible to use the lookup module to get the custom field choices.

_What are the drawbacks of this change?_

I do not know what happens when someone using it with an NetBox version lower than v3.6.0.

_Is it backwards-compatible?_

NetBox <v3.6.0; Use `custom-fields` to get the choices.
Netbox >v3.6.0; Use `custom-field-choice-sets`

...

Added `custom-field-choice-sets` to the lookup plugin.

## Double Check

### DIFF
--- a/plugins/lookup/nb_lookup.py
+++ b/plugins/lookup/nb_lookup.py
@@ -181,6 +181,7 @@ def get_endpoint(netbox, term):
         "console-server-ports": {"endpoint": netbox.dcim.console_server_ports},
         "content-types": {"endpoint": netbox.extras.content_types},
         "custom-fields": {"endpoint": netbox.extras.custom_fields},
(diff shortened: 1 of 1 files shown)


sc68cal (approved):

sc68cal (changes requested):
You need to make this change in a way that it supports NetBox versions less than 3.6 since we still support those per our policy

sc68cal (approved):

PR #1202, opened 2024-04-05
Added label to power port module

sapergus:
## Related Issue

No related issues.

## New Behavior

Added 'label' as a none required option to netbox_power_port module.
...

Possible to add  a label to a power port in NetBox.
...

We use NetBox in my organisation and I need to be able to create labels for powerports.
Since the option to add a label is optional it should not be a problem with backward compatibility.
...

Added a description of the label option in plugins/modules/netbox_power_port.py and generated a new version of
docs/plugins/netbox_power_port_module.rst.
...

...

## Double Check

### DIFF
--- a/docs/plugins/netbox_power_port_module.rst
+++ b/docs/plugins/netbox_power_port_module.rst
@@ -6,14 +6,33 @@
 .. |antsibull-internal-nbsp| unicode:: 0xA0
     :trim:
 
-.. meta::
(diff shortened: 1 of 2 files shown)


sc68cal (changes requested):
The markdown docs are automatically generated I believe, so your change should only just be to the `netbox_power_port.py` file.

sc68cal (approved):

PR #1081, opened 2023-10-05
Add out of band IP option

Shourai:
Add oob_ip documentation

## New Behavior
Add the option to use the out of band IP as `ansible_host` instead of primary_ip.

## Contrast to Current Behavior
There is no option to use the out of band ip currently

## Discussion: Benefits and Drawbacks
Allow users to use the new out of band ip option in their playbooks.
Also the query filter `has_oob_ip: True` is available. 

## Proposed Release Note Entry
Add the ability to set out of band IP as ansible host.

## Double Check

* [x] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netbox-community/ansible_modules/blob/devel/CONTRIBUTING.md).
* [x] I have explained my PR according to the information in the comments or in a linked issue.
* [x] My PR targets the `devel` branch.

PR #1211, opened 2024-04-15
feat: make `oob_ip` available regardless of `oob_ip_as_primary_ip`

akire0ne:
## Related Issue

Related to https://github.com/netbox-community/ansible_modules/pull/1081

## New Behavior

Makes `oob_ip` always available in the hostvars if it exists.

Not sure why we make `oob_ip` available only when `oob_ip_as_primary_ip` is set.

I believe there is value to access this property without altering `ansible_host`.

None

- `oob_ip` is now returned in hostvars if set
- Add the ability to set out of band IP as ansible host.

## Double Check

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -1875,8 +1875,10 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             self.inventory.set_variable(hostname, "primary_ip6", extracted_primary_ip6)
 
         extracted_oob_ip = self.extract_oob_ip(host=host)
-        if extracted_oob_ip and self.oob_ip_as_primary_ip:
-            self.inventory.set_variable(hostname, "ansible_host", extracted_oob_ip)
+        if extracted_oob_ip:
+            self.inventory.set_variable(hostname, "oob_ip", extracted_oob_ip)
+            if self.oob_ip_as_primary_ip:
+                self.inventory.set_variable(hostname, "ansible_host", extracted_oob_ip)
 
         for attribute, extractor in self.group_extractors.items():
             extracted_value = extractor(host)


sc68cal:
@Shourai can you please take a look, since you implemented the original feature?

Shourai:
I have checked it out and tested it.
It's a really nice solution.

- With `oob_ip_as_primary_ip: false` and oob_ip defined but primary ip undefined, the variable `oob_ip` will be available but `ansible_host` will not be.

- With `oob_ip_as_primary_ip: true` and oob_ip defined but primary ip undefined, the variable `oob_ip` will be available and `ansible_host` will also be available.

It's useful if you have `oob_ip` defined and want to use that variable but not overwrite the `primary_ip` as `ansible_host`.

sc68cal:
That is a large matrix of behaviors. I don't know if I am a fan of this change.

sc68cal:
Frankly I would rather have one configuration setting, where you either use the `primary_ip` of a device, or the `oob_ip` of a device, and leave it at that.

sc68cal (changes requested):
Creates too complex behaviors

akire0ne:
Folks, two different things here
- `oob_ip` is displayed on the Device UI in the same manner as `Primary IPv4` and `Primary IPv6` - So for me I would expect to be displayed in my hostvars, in the same fashion as `primary_ip4` and `primary_ip6`are.
- Ability to set `oob_ip` as a `ansible_host` - We don't have that need, but as @Shourai already implemented something with `oob_ip` so I didn't wanted to clash with this.

Whatever works for me as long `oob_ip` is available when set on the device. 

@sc68cal are you suggesting something like an `ansible_host_ip_source` setting, which could be set to either `primary_ip4` (default) or `oob_ip` ? 

FYI our business case is to be able to fetch / identify OOBM / IPMI easily for our devices.

sc68cal (approved):
OK. It took me a couple reads through your comments @akire0ne and @Shourai, because for some reason my brain was just not working properly and I was getting confused.

It was not because of anything you wrote, but for some reason it just didn't stick with me for some reason. 

It's sad because the code is quite clear what is going on but I just wasn't getting it. Sorry. 

This is good to go, thank you for the contribution and your patience, I will try and do better next time.

akire0ne:
Thanks @sc68cal  ! Very appreciated

PR #1273, opened 2024-06-19
Add ability to rename host inventory variables using patterns

href:
## Related Issue

None.

## New Behavior

Certain environments may have conflicts with the variables given by Netbox, and the variables used in the existing Ansible codebase.

With this change, such cases can be worked around by renaming individual variables, or whole groups of them.

For example, this will rename all `cluster*` variables to have a `netbox__` prefix instead (e.g., `cluster_group` -> `netbox__cluster_group`):

```yaml
rename_variables:
  - pattern: 'cluster(.*)'
    repl: 'netbox__cluster\1'
```

Uses a list, instead of a dict, to ensure that the order of evaluation is strictly defined across all Python versions, and to add the ability to exclude certain variables from being rewritten. For example:

```yaml
rename_variables:
  # Keep cluster_type the same
  - pattern: 'cluster_type'
    repl: 'cluster_type'

  # Rename all other cluster* variables
  - pattern: 'cluster(.*)'
    repl: 'netbox__cluster\1'
```

Currently, the variable names are hard-coded and cannot be modified.

With this change, it becomes easier to adapt Netbox to code-bases where some of the variables found in Netbox already have a meaning.

The change is purely optional, so anyone that does not need it, can just ignore it.

The documentation is updated in the module. I assume this will be published once merged.

Add ability to rename host inventory variables using patterns, to avoid conflicts with existing variables.

...

## Double Check

### DIFF
--- a/changelogs/fragments/1273-ability-to-rename-host-inventory-vars.yml
+++ b/changelogs/fragments/1273-ability-to-rename-host-inventory-vars.yml
@@ -1,0 +1,2 @@
+minor_changes:
+      - Add ability to rename variables set on the host by ``netbox.netbox.nb_inventory`` through configuration.
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -249,6 +249,16 @@ DOCUMENTATION = """
             description: Use out of band IP as `ansible host`
             type: boolean
             default: false
+        rename_variables:
+            description:
+                - Rename variables evaluated by nb_inventory, before writing them.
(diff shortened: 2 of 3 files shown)


rodvand:
This looks like a sensible change - could you:
- make sure the linting checks pass (run black on your code)
- add a changelog fragment https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment
- add some tests, or at least suggest some test cases we can implement

href:
Sure, let me get back to you on that.

href:
Please take another look, I've got the tests to pass and added a simple unit test.

Issue #1300, opened 2024-08-15
[Feature]: nb_inventory - allowing api_endpoint to be a variable


bookshark13:
### NetBox version

v3.19.1

### Feature type

Change to existing Plugin

### Proposed functionality

Not a big change to code if you add:
`self.api_endpoint = self.templar.template( self.get_option("api_endpoint"), fail_on_undefined=False)`
to the nb_inventory plugin [at](https://github.com/netbox-community/ansible_modules/blob/16d86773e8bc1cc767daaf6cd632fa03ca24c73e/plugins/inventory/nb_inventory.py#L1597)
this will allow for the api_endpoint the option of being a variable all while maintaining it's ability to be a string. 

### Use case

In current Netbox Module the api_endpoint variable can only be a string. Ex: "https://netbox.at.your.domain.com"
If you try to inject a variable for api_endpoint like so `api_endpoint: "{{ lookup('ansible.builtin.env', 'NETBOX_URL') }}"`
The plugin will output that _"{{ lookup('ansible.builtin.env', 'NETBOX_URL') }}"/api/status_  is not valid.

If you manage multiple environments with possibly different domain names this is quite useful. It allows you to define once in the code and carry it over to multiple environments while maintaining a credential with things like Ansible Controller vs changing your code.

### External dependencies

N/A

PR #1301, opened 2024-08-15
Issue #1300 - nb_inventory injecting a variable as api_endpoint

bookshark13:
Fixes issue-#1300
allows api_endpoint to be a variable and string

## Related Issue
#1300 

## New Behavior

calls the templar.template module to transform the api_endpoint if referenced as a variable e.g: "{{ fake_var }}"

_fake_var: "https://netbox.test.domain.com"_
Current behavior for a inventory file is referencing `api_endpoint: "{{ fake_var }}"` 
for the above example you would receive an error to the affect of 

`[WARNING]:  * Failed to parse
<location of your netbox plugin> with auto plugin:
unknown url type: "{{ fake_var }}/api/status"`

this solves that issue and the result is no error and we can see 
`Fetching: https://netbox.test.domain.com/api/status`

- This could be beneficial to the plugin as a whole, not everyone has one netbox environment. As well nor do they share domain names. This change allows for dynamically calling the api_endpoint. While also not changing anyones current setup
- I believe there are no real drawbacks for this change. It does not take away functionality. Truly it only increases the way you can call the api_endpoint
- Yes it is backwards-compatible

N/A

netbox.nb_inventory: api_endpoint option can now take variable calls E.G: `api_endpoint: "{{ netbox_url }}" `

## Double Check

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -1594,7 +1594,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         except Exception:
             cached_api_version = None
             cache = None
-
+        self.api_endpoint = self.templar.template( self.get_option("api_endpoint"), fail_on_undefined=False)
         status = self._fetch_information(self.api_endpoint + "/api/status")
         netbox_api_version = ".".join(status["netbox-version"].split(".")[:2])
 


sc68cal:
The netbox inventory plugin already supports the environment variable `NETBOX_API` as [documented here](https://netbox-ansible-collection.readthedocs.io/en/latest/plugins/environment_variables.html#envvar-NETBOX_API).

Can you shed some light on why this change is required when we already have this environment variable?

sc68cal (changes requested):
`nb_inventory` plugin already supports an environment variable, please see the existing documentation.

There has been discussion about the Ansible _modules_ being able to use the environment variable, which both your issue and #1363 discusses, but this code only touches the `nb_inventory` plugin which already has support.

Woefdram:
This would really help. Setting an environment variable is a primitive solution if you work with several people in several environments.

Being able to specify the netbox_api somewhere else would solve a lot of hassle for me.

andrewleininger:
Another use case for templar here is lookup plugins pulling the API endpoint value from a secrets manager or parameter store (community.general.onepassword, amazon.aws.ssm_parameter, community.general.consul_kv, etc.).

I'd like to be able to do:

```yaml
---
# api_endpoint: https://my.netbox.fqdn
api_endpoint: "{{ lookup('community.general.onepassword', 'My NetBox Item', vault='My Vault', section='API', field='endpoint') }}"
token: "{{ lookup('community.general.onepassword', 'My NetBox Item', vault='My Vault', section='API', field='ro-token') }}"
```

Issue #975, opened 2023-04-09
[Bug]: Cannot assign existing IP to network interface


cfiehe:
### Ansible NetBox Collection version

v3.12.0

### Ansible version

```bash
ansible [core 2.14.1]
  config file = /home/ansible/automation/projects/provisioning/ansible.cfg
  configured module search path = ['/home/ansible/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ansible/.local/lib/python3.10/site-packages/ansible
  ansible collection location = /home/ansible/automation/projects/provisioning/collections
  executable location = /home/ansible/.local/bin/ansible
  python version = 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0] (/usr/bin/python3)
  jinja version = 3.0.3
  libyaml = True
```

### NetBox version

v3.4.7

### Python version

3.10

### Steps to Reproduce

If an IP address already exists in our Netbox, it cannot be assigned to an interface. The module always instructs the Netbox to create a new IP. This results in a `Duplicate IP address` exception, when `Enforce unique space` is enabled, or in a duplicate IP.

```
- name: Create an IP address
  netbox.netbox.netbox_ip_address:
    netbox_url: "{{ netbox_url }}"
    netbox_token: "{{ netbox_token }}"
    data:
      address: 10.1.60.232/24
      tenant: my_company
      vrf: my_company-RFC1918
      status: Active
    state: present

- name: Assign IP address to existing interface
  netbox.netbox.netbox_ip_address:
    netbox_url: "{{ netbox_url }}"
    netbox_token: "{{ netbox_token }}"
    data:
      address: 10.1.60.232/24
      assigned_object:
        virtual_machine: any_vm.local
        name: ens224
      tenant: my_company
      vrf: my_company-RFC1918
      status: Active
    state: present
```

### Expected Behavior

The existing IP should be modified and should be attached to the specified interface.

### Observed Behavior

The VRF `my_company-RFC1918` uses the option `Enforce unique space `, that is why the assignment fails with a `Duplicate IP address`, because the module tries to create a new IP instead of modifying the existing one.

Zwordi:
Hello,

My 2cents here as I’m looking of kind of the same topic.
The "Enforce unique space" is the good way. As by having an unique address you’re being certain that this one will not be treated differently.
The need here is about being able to update an ip record with the "state:present" attribute.

d-zalewski:
[...]

PR #1323, opened 2024-09-20
fixing bug #975

Ido-Don:
## Related Issue

The issue this PR is trying to fix is #975 

## New Behavior

this pull request changes the data object to include the assigned_object_id and assigned_object_type values.
it takes the interface_id that exists in the interface value and adds an assigned_object_id and an assigned_object_type item.

the current behavior will not assign an IP to the interface.

This bug prevents the assigning of an IP to an interface.
it is backwards compatible.

It is worth noting that there is a duplication between the assigned_object and interface values.
In my opinion it is best to deprecate the interface value internally and make people use only the assigned_object value.

the documentation already states this functionality, it was not implemented but it should have.

netbox_ip_address - Fixed the problem preventing assignment of an IP address to a network interface

## Double Check


sc68cal:
This looks good to me, just needs a changelog put into the `fragments` directory and then maybe a test to verify the behavior?

Ido-Don:
I added the tests and fragment to the PR.
Should i open an issue so we could discuss the deprecation of the interface value?

sc68cal (approved):
Thank you for the contribution!

sc68cal:
>  Should i open an issue so we could discuss the deprecation of the interface value?

Yes please do

Ido-Don:
> > Should i open an issue so we could discuss the deprecation of the interface value?
> 
> Yes please do

I opened issue #1338, you are welcome to come and take a look 😁

Ido-Don:
> Thank you for the contribution!

Thank you for reviewing my code!
now all that i need is to marge the changes 😄

PR #1348, opened 2024-11-02
Add missing elements to netbox_fhrp_group

Hramoff:
## New Behavior

I can't specify a virtual IP when creating an FHRP group via Ansible, so I added this functionality

Two new keys `ip_address` and `ip_status` have been added to netbox_fhrp_group.py

My solution is backwards compatible because I use `required=False` and user tasks will not stop working.
Next. This functionality is needed to conveniently manage FHRP groups.

As far as I know, the documentation in the wiki is automatically generated from the `DOCUMENTATION` variable in the module, so I added a description there of what I did and how to work with it

Add missing elements to netbox_fhrp_group

### DIFF
--- a/plugins/modules/netbox_fhrp_group.py
+++ b/plugins/modules/netbox_fhrp_group.py
@@ -78,6 +78,22 @@ options:
         required: false
         type: list
         elements: raw
+      ip_address:
+        description:
+          - Virtual IP address for the FHRP group
+        type: str
+        required: false
+      ip_status:
+        description:
+          - IP address status
+        choices:
+          - dhcp
+          - reserved
+          - deprecated
+          - slaac
+          - active
+        type: str
+        required: false
       custom_fields:
         description:
           - Must exist in NetBox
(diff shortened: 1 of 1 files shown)


Hramoff:
I confused how additional ones are added.

I wanted to add functionality so that the IP address could be added to the FHRP group. I thought that the keys from the json that is sent to the web interface coincide with the keys in Ansible, but as it turns out, they do not.


I looked at what the debug outputs:

```yaml
    - name:
      debug:
        msg: "{{ query('netbox.netbox.nb_lookup', 'ip-addresses', api_endpoint=netbox_url, token=netbox_user_token, api_filter=filter_description)  }}"
```

```json
        {
            "key": 11186,
            "value": {
                "address": "10.22.5.88/32",
                "assigned_object": {
                    "description": "demo-network",
                    "display": "keepalived-vip VRRPv3: 3 (10.22.5.88/32)",
                    "group_id": 3,
                    "id": 30,
                    "protocol": "vrrp3",
                    "url": "https://netbox.local/api/ipam/fhrp-groups/30/"
                },
                "assigned_object_id": 30,
                "assigned_object_type": "ipam.fhrpgroup",
        }
}

```

I saw that in order to link an address to a fhrp group you need to specify its ID and type, respectively `assigned_object_id` and `assigned_object_type`

I decided to add these two parameters because I think they will be universal for other types and IDs.

Hramoff:
I think that it is better to search using the name key rather than group_id.

Issue #1401, opened 2025-03-24
[Bug]: netbox_circuit_termination fails with "A circuit termination must attach to a terminating object."


nsthompson:
### Ansible NetBox Collection version

v3.21.0

### Ansible version

```bash
ansible [core 2.18.3]
  config file = None
  configured module search path = ['/home/vscode/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.13/site-packages/ansible
  ansible collection location = /home/vscode/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.13.2 (main, Feb 25 2025, 05:25:21) [GCC 12.2.0] (/usr/local/bin/python3.13)
  jinja version = 3.1.6
  libyaml = True
```

### NetBox version

v4.2.5

### Python version

3.11

### Steps to Reproduce

The tasks I am running look like this:

```yaml
- name: Configure A-side Circuit Terminations for Circuit {{ circuit.data.id }} from {{ circuit.data.provider }}
  netbox.netbox.netbox_circuit_termination:
    data:
      circuit: "{{ circuit.data.id }}"
      description: "{{ circuit.terminations.a.description | default(omit) }}"
      mark_connected: "{{ circuit.terminations.a.mark_connected | default(omit) }}"
      port_speed: "{{ circuit.terminations.a.port_speed | default(omit) }}"
      pp_info: "{{ circuit.terminations.a.pp_info | default(omit) }}"
      provider_network: "{{ circuit.terminations.a.provider | default(omit) }}"
      site: "{{ circuit.terminations.a.site | default(omit) }}"
      term_side: A
      upstream_speed: "{{ circuit.terminations.a.upstream_speed | default(omit) }}"
      xconnect_id: "{{ circuit.terminations.a.xconnect_id | default(omit) }}"
    state: "{{ circuit.terminations.a.state }}"
  when: circuit.terminations.a is defined

- name: Configure Z-side Circuit Terminations for Circuit {{ circuit.data.id }} from {{ circuit.data.provider }}
  netbox.netbox.netbox_circuit_termination:
    data:
      circuit: "{{ circuit.data.id }}"
      description: "{{ circuit.terminations.z.description | default(omit) }}"
      mark_connected: "{{ circuit.terminations.z.mark_connected | default(omit) }}"
      port_speed: "{{ circuit.terminations.z.port_speed | default(omit) }}"
[...]

PR #1402, opened 2025-03-24
Updated netbox_circuit_termination.py to support NetBox 4.2.x

nsthompson:
## Related Issue

#1401 

## New Behavior

This fix is to resolve the changes created by the [NetBox 4.2.0 release](https://github.com/netbox-community/netbox/releases/tag/v4.2.0) relating to:

`The site and provider_network foreign key fields on circuits.CircuitTermination have been replaced by the termination generic foreign key.`

Current behavior results in module failure with the error:

`A circuit termination must attach to a terminating object.`

This fix resolves the issue with NetBox 4.2.x.  It is backward compatible with prior versions.

Documentation in the module has been updated.

Updated netbox_circuit_termination.py to support NetBox 4.2.x breaking changes.

## Double Check

* [ x ] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netbox-community/ansible_modules/blob/devel/CONTRIBUTING.md).
* [ x ] I have explained my PR according to the information in the comments or in a linked issue.
* [ x ] My PR targets the `devel` branch.

### DIFF
--- a/changelogs/fragments/netbox_circuit_termination.yml
+++ b/changelogs/fragments/netbox_circuit_termination.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - netbox_circuit_termination - Add parameters termination_id and termination_type for NetBox 4.2+
--- a/plugins/modules/netbox_circuit_termination.py
+++ b/plugins/modules/netbox_circuit_termination.py
@@ -51,13 +51,15 @@ options:
         version_added: 3.5.0
       termination_id:
         description:
-          - The ProviderNetwork, Location, Site, Region, or SiteGroup ID of the circuit termination will be assigned to.  This is used with NetBox versions >= 4.2.0.
+          - The ProviderNetwork, Location, Site, Region, or SiteGroup ID of the circuit termination will be assigned to.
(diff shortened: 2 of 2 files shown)


sc68cal:
This needs to be done in a way that preserves backwards compatibility with older NetBox versions since we still use those for CI tests, and we have by convention supported the two most recent releases, which at this point would be `4.2` and `4.1`

nsthompson:
> This needs to be done in a way that preserves backwards compatibility with older NetBox versions since we still use those for CI tests, and we have by convention supported the two most recent releases, which at this point would be `4.2` and `4.1`

Thanks for the feedback.  I just pushed new changes that preserve backwards compatibility and prevent parameter overlap between the params required for versions pre 4.2.0 and those for versions 4.2.0 and later.

The fix was tested against NetBox Cloud running 4.1.11 as well as 4.2.5.

sc68cal:
Looks good, just some small style changes for `black` to be happy, and then a changelog fragment and then I think this is good to go

sc68cal:
Doc build failure appears unrelated to your changes.

sc68cal (approved):

Issue #1195, opened 2024-03-19
nb_inventory exception after fetching /api/status instead of results/devices


double-p:
### Ansible NetBox Collection version

v3.17.0

### Ansible version

```bash
ansible [core 2.16.3]
  config file = /Users/pbuehler/Workspace/dns-ansible/ansible.cfg
  configured module search path = ['/Users/pbuehler/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/9.2.0/libexec/lib/python3.12/site-packages/ansible
  ansible collection location = /Users/pbuehler/Workspace/galaxy
  executable location = /usr/local/bin/ansible
  python version = 3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 (clang-1500.1.0.2.5)] (/usr/local/Cellar/ansible/9.2.0/libexec/bin/python)
  jinja version = 3.1.3
  libyaml = True
```

### NetBox version

v3.7.4

### Python version

3.11

### Steps to Reproduce

Just run ansible-inventory with this minimal YML to try getting results:
```bash
user$ cat nv_inv.yml
plugin: netbox.netbox.nb_inventory
api_endpoint: http://localhost:8888/netbox
token: xxxx
user$
user$ ansible-inventory --list -i  nb_inv.yml
{
    "_meta": {
        "hostvars": {}
    },
    "all": {
        "children": [
            "ungrouped"
        ]
    }
}
```

### Expected Behavior

return devices (as it does when just calling the API with curl)

### Observed Behavior

Looking into -vvvv I can see this:
```bash
Fetching: http://localhost:8888/netbox/api/status
toml declined parsing /Users/pbuehler/Workspace/dns-ansible/nb_inv.yml as it did not pass its verify_file() method
[WARNING]:  * Failed to parse /Users/pbuehler/Workspace/dns-ansible/nb_inv.yml with auto plugin: '/api/dcim/devices/'
  File "/usr/local/Cellar/ansible/9.2.0/libexec/lib/python3.12/site-packages/ansible/inventory/manager.py", line 293, in parse_source
    plugin.parse(self._inventory, self._loader, source, cache=cache)
  File "/usr/local/Cellar/ansible/9.2.0/libexec/lib/python3.12/site-packages/ansible/plugins/inventory/auto.py", line 59, in parse
    plugin.parse(inventory, loader, path, cache=cache)
[...]

PR #1407, opened 2025-04-13
inventory plugin: support netbox endpoints running in a non-root context path (subpath)

lanefu:
## Related Issue

#1408 

## New Behavior

Updated Plugin to parse the full path of the netbox server and compose API endpoints accordingly.
...

## allow inventory to support netbox endpoints on host.example.com/netboxpath/

This is a patch I originally made for v3.13 which had updated parsing of the openapi endpoint and assumes that netbox is always running in the root context rather than a path.     This performs the additional url parsing to derive the correct inventory API endpoints if netbox is running on a path rather than a root context.

I've tested this rebased patch on latest netbox running on in the root context path, as well as my legacy deployment running on host.com/netbox/
...

- allows users to run in additional supported configurations ex: somehost.example.com/netbox/ rather than only netbox.example.com
-

...

...

update inventory plugin to support netbox API running on non-root context paths.
...

## Double Check

### DIFF
--- a/changelogs/fragments/nb_inventory.yml
+++ b/changelogs/fragments/nb_inventory.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - nb_inventory - Fixed empty inventory results when netbox server URL is a non-root path
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -406,6 +406,8 @@ from ansible.module_utils._text import to_text, to_native
 from ansible.module_utils.urls import open_url
 from ansible.module_utils.six.moves.urllib import error as urllib_error
 from ansible.module_utils.six.moves.urllib.parse import urlencode
+from ansible.module_utils.six.moves.urllib.parse import urlparse
+
 from ansible.module_utils.six import raise_from
 
 try:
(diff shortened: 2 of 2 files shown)


lanefu:
Also related to #1195

lanefu:
For some additional background in response to #1195 

When I originally implemented this fix a year ago, I had already been using netbox inventory and ipam modules for quite a while (using a subpath)  The inventory plugin broke when some changes were made in response to querying OpenAPI.

However, from what I recall, it only broke the inventory plugin... The other modules relying on the netbox_util module already had better api path parsing and it just worked.   My solution was inspired from what was in netbox_util

https://github.com/netbox-community/ansible_modules/blob/cd0b654138d018ae2beab4810e69138d5b5d8142/plugins/module_utils/netbox_utils.py#L857

sc68cal:
The integration test failure against 4.2 is unrelated to this change (it's an issue with how NetBox reports its versions in that branch, it's a regression), and the doc build failure is unrelated to this change as well. I think this is good to go.

sc68cal (approved):

leitwerk-ag:
Oh, I'd love to see this one merged 😍

sc68cal:
@rodvand I'm planning on merging this regardless of the CI failure, which is unrelated to this change. Thoughts?

PR #1416, opened 2025-05-16
fix: nb_lookup plugin: Allow mac-addresses lookup

shaderecker:
## New Behavior

Add possibility to query mac-addresses with the nb_lookup plugin for NetBox 4.2+

...

Feature is missing so far

...

- mac-addresses was added in Netbox 4.2 so we need to follow the feature set
- No drawbacks
- The mac-addresses feature wasn't available in Netbox versions lower than 4.2 so it won't impact those.

...

n/a

...

nb_lookup plugin - Allow mac-addresses lookup

...

## Double Check


sc68cal (changes requested):
I believe this needs to check the API version before adding to `netbox_endpoint_map` - because we still support `4.1` where this API is not available.

There are lots of examples of this, below the line that you have added.

sc68cal (approved):

PR #1421, opened 2025-05-22
Update netbox_service.py

garnser:
# Add support for `parent_object_type` and `parent_object_id` in `netbox_service.py`

## New Behavior

* Extends the **NetBox Service** module with two new required parameters introduced in NetBox 4.3:

| Parameter            | Type | Required | Since |
|----------------------|------|----------|-------|
| `parent_object_type` | str  | yes      | 3.22  |
| `parent_object_id`   | int  | yes      | 3.22  |

* Allows users to bind a service directly to **any** NetBox object type (device, VM interface, IP address, etc.) by specifying its type and primary‑key ID.

* Prior to NetBox 4.3 the API accepted `device`, `virtual_machine`, or similar fixed keys.  
* Without these new fields, the module now fails with **“A parent object (type & id) must be specified”** errors when talking to NetBox ≥ 4.3.

### Benefits
* Restores full compatibility with the current NetBox API.
* Change is additive, so it remains harmless for NetBox ≤ 4.2 (fields are ignored by the API).

### Drawbacks
* Slightly expands the module interface; playbooks **must** supply the new params when upgrading to NetBox 4.3+.

```text
* **netbox_service** – add required NetBox 4.3 fields
  * new parameters: `parent_object_type` (str), `parent_object_id` (int)
```

## Double Check

* [X] My PR targets the **devel** branch.

### DIFF
--- a/plugins/modules/netbox_service.py
+++ b/plugins/modules/netbox_service.py
@@ -85,6 +85,18 @@ options:
           - Must exist in NetBox and in key/value format
         required: false
         type: dict
+      parent_object_type:
+        description:
+          - Specifies which object type to associate with service.
+        required: true
+        type: str
+        version_added: "3.22.0"
+      parent_object_id:
+        description:
+          - Specifies which object to associate with service.
+        required: true
+        type: int
+        version_added: "3.22.0"
     required: true
 """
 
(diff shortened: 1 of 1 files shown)


sc68cal (changes requested):
This needs to handle the case where the NetBox server is a 4.1 version, similar to how we handle other API comparability issues.

yodzeb:
Hi,

Just seen this PR for the same issue i was facing.
I made https://github.com/netbox-community/ansible_modules/pull/1427 for fixing, which i think might be more user-friendly without exposing low-level parameters to the ansible module user.

wdyt ?

PR #1427, opened 2025-06-12
fixes #1426 : broken netbox_service module

yodzeb:
## Related Issue

https://github.com/netbox-community/ansible_modules/issues/1426

## New Behavior

netbox_service module is working again, with either ***device*** or ***virtual_machine*** parameter, corretly mapped to API parameters

## Discussion: Benefits and Drawbacks

Not sure if the extra params should be cleaned in netbox_ipam.py after being moved to API parameters., but it's working.

## Changes to the Documentation

None

## Proposed Release Note Entry

Changelog fragment pushed
## Double Check

* [X] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netbox-community/ansible_modules/blob/devel/CONTRIBUTING.md).
* [X] I have explained my PR according to the information in the comments or in a linked issue.
* [X] My PR targets the `devel` branch.

PR #1438, opened 2025-07-07
[WIP] Fix netbox_inventory service query

StoffelCPR:
## Related Issue

#1430 

## New Behavior

This is a first iteration of a possible fix for the service integration to post 4.3 Netbox versions.

...

## Contrast to Current Behavior

Services remain empty as Netbox has introduced breaking changes

...

## Discussion: Benefits and Drawbacks

Benefit:

re-established function to fetching services

Drawback:

Dumb quick fix. I'd need some help properly onboarding my changes into your change management

...

## Changes to the Documentation

Probably no

...

## Proposed Release Note Entry

Fixed service mapping to new Netbox parent object structure

...

## Double Check

* [ ] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netbox-community/ansible_modules/blob/devel/CONTRIBUTING.md).
* [ x] I have explained my PR according to the information in the comments or in a linked issue.
* [ x] My PR targets the `devel` branch.

PR #1439, opened 2025-07-08
Fix sevral integration tests failure

benoit-ferre:
Fix sevral bugs dicovered by integration tests failure

## Related Issue
N/A
I did not find any open issue and I fix tem directly without fill a bug.
- Fix service creation with netbox prior of 4.3.0 (since #1427 I think)
- Use dedicated function to check netbox version istead of self.full_version for rack.
- Fix typos in tag integration tests.
- Fix integration test for circuit termination, missing assignment
- Fix integration test for service
  
## New Behavior
Only bug fix

N/A

Bug fix

N/A

- Add netbox version check to support service creation for netbox version prior of 4.3

## Double Check

### DIFF
--- a/changelogs/fragments/1439-fix-sevral-bug-discovered-by-integration-tests.yml
+++ b/changelogs/fragments/1439-fix-sevral-bug-discovered-by-integration-tests.yml
@@ -1,0 +1,6 @@
+bugfixes:
+  - Add netbox version check to support service creation for netbox version prior of 4.3
+  - Use dedicated function to check netbox version istead of self.full_version for rack.
+  - Fix typos in tag integration tests.
+  - Fix integration test for circuit termination, missing assignment
+  - Fix integration test for service


StoffelCPR:
Hey,

If you want take a look into the integration test targets 4.2 and 4.3. Technically they're failing. You just can't see it because the printout of the JSON debug output fails and the exit(1) is never reached.

I'm trying to fix that in my pull request but I've just started looking at this source code. If you're faster that'd be greatly aprpeciated. #1438

benoit-ferre:
Hello @StoffelCPR ,
I saw there is a problem with service since version 4.2.
But I was only focused in making unit tests passing again in this PR.
For unit services tests I think my solution is more a workaround than a fix. I did not analyse so much. 
You can open a new discussion an reference it on your PR so we can dicuss of this in there.

Issue #1455, opened 2025-08-23
[Bug]: Changing order of permissions causes unnecessary update


dchiquito:
### Ansible NetBox Collection version

v3.10.0

### Ansible version

```bash
ansible [core 2.18.8]
  config file = /home/daniel/git/ansible_modules/ansible.cfg
  configured module search path = ['/home/daniel/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/daniel/.local/lib/python3.13/site-packages/ansible
  ansible collection location = /home/daniel/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/daniel/.local/bin/ansible
  python version = 3.13.3 (main, Apr 22 2025, 00:00:00) [GCC 15.0.1 20250418 (Red Hat 15.0.1-0)] (/usr/bin/python3)
  jinja version = 3.1.6
  libyaml = True
```

### NetBox version

v3.4.0

### Python version

3.11

### Steps to Reproduce

There is an existing user with existing permissions `['foo', 'bar']`, and a playbook like this is run:
```
- name: "Update user"
  netbox.netbox.netbox_user:
    data:
      username: MyUser
      permissions:
        - bar
        - foo
    state: present
```

### Expected Behavior

The user will not be updated, because the permission sets are equivalent.

### Observed Behavior

The user is updated, because the permission lists are considered different.

PR #1456, opened 2025-08-23
Use set comparison for updating permissions, groups, actions, and object types

dchiquito:
## Related Issue
https://github.com/netbox-community/ansible_modules/issues/1455

## New Behavior

`user.groups`, `user.permissions`, `user_group.permissions`, `permission.actions`, and `permission.object_types` are all serialized as lists. When compared as lists, however, any ordering changes means that the `user`, `user_group`, or `permission` are updated unnecessarily. 

Convert these fields into sets prior to comparison during the update process.

Unnecessary updates can be very costly for objects with large numbers of permissions, so avoiding that would be great.

The only drawback I can think of would be a negligible increase in compute, which is offset by fewer required network requests.

I don't think ordering of these fields is a feature so this would be backwards compatible.

No changes.

* `user.groups`, `user.permissions`, `user_group.permissions`, `permission.actions`, and `permission.object_types` are now treated as unordered sets for update comparison purposes.

## Double Check

### DIFF
--- a/tests/integration/targets/v4.0/tasks/netbox_permission.yml
+++ b/tests/integration/targets/v4.0/tasks/netbox_permission.yml
@@ -211,6 +211,4 @@
   ansible.builtin.assert:
     that:
       - not test_nine['changed']
-      # actions seem to be ordered randomly so we cannot test them here
-      - test_nine['permission']['object_types'][0] == 'vpn.tunneltermination'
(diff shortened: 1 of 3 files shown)


sc68cal:
Working on a fix to devel that hopefully will fix the sanity failures

sc68cal (approved):
Looks good. Also, thank you for the well written commit messages, it answered a few questions I had in review, saving time.

PR #1487, opened 2025-11-13
Convert more keys/attribute values from list to set

### LINKED ISSUE #1486: [Bug]: Updating Permissions always lists object_types to be changed

v3.21.0

### Ansible version

```bash
ansible [core 2.15.13]
```

When using `netbox.netbox.netbox_permission` to update given permissions, the field `object_types` always gets return as changed, cause the list is compared as list, not set. The order of the object_types it applies to are not important.

This is similar to `tags`

I would propose to
a) bring this in line with the `pynetbox` response (which defines fields are sets)
b) define the field `object_types` as set as well (pynetbox as well)

### NetBox version

v4.4.6

### Python version

3.9

### Steps to Reproduce

```
- name: update permissions
  netbox.netbox.netbox_permission:
    netbox_url: ENDPOINT_URL
    netbox_token: API_TOKEN
    data:
      name: example-permission
      enabled: true
      actions: ['view', 'add', 'change', 'delete']
      object_types:
        - dcim.interface
        - dcim.module
        - dcim.device
      constraints: {}
```

Depending on the response and your given order, ansible always detects a change.

### Expected Behavior

When the list of object_types the permissions applies to doesn't change, ansible should not report that something has changed.

### Observed Behavior

```
[...]

changed: [netbox-vm] => (item=ipam-perms)
ok: [netbox-vm] => (item=ipam-perms)
--- before
+++ after
@@ -1,8 +1,8 @@
     object_types:
+    - ipam.ipaddress
+    - ipam.prefix
     - ipam.vrf
-    - ipam.prefix
-    - ipam.iprange
-    - ipam.ipaddress
     - ipam.vlangroup
     - ipam.vlan
+    - ipam.iprange

[...]
```

As you see, the order ist always random (iprange > vrf ....)

dhoffend:
## Related Issue

Fixes #1486 

## New Behavior

In addition to `tags`, the fields `tagged_vlans` and `object_types` should be considered a set, not an ordered list. This should be in sync with `pynetbox` as well. (https://github.com/netbox-community/pynetbox/blob/master/pynetbox/core/response.py#L25-L26) I'll propose that object_types will be added there as well.

This is a follow up on #1456 to take a more generic approach when lists should be converted to/compared as set

...

The field `object_types` is currently handled as a sorted list, but the response is unsorted resuting in constand changed tasks, even nothing has changed.

...

I don't see any drawbacks as object_types is not considered to be ordered.

...

Nothing, as it doesn't affect the outcome, it just fixes a display issue (changed vs ok)

Take a more generic approach for set comparison. Other models have object_types too
...

## Double Check

### DIFF
--- a/changelogs/fragments/1486-fix-list-as-set.yml
+++ b/changelogs/fragments/1486-fix-list-as-set.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - Take a more generic approach for set comparison. Other models have object_types too
--- a/plugins/module_utils/netbox_users.py
+++ b/plugins/module_utils/netbox_users.py
@@ -15,9 +15,6 @@ NB_PERMISSIONS = "permissions"
 NB_TOKENS = "tokens"
 NB_USERS = "users"
 
-# These suboptions are lists, but need to be modeled as sets for comparison purposes.
-LIST_AS_SET_KEYS = set(["permissions", "groups", "actions", "object_types"])
-
 
 class NetboxUsersModule(NetboxModule):
     def __init__(self, module, endpoint):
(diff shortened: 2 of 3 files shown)


sc68cal:
Please rebase this on the latest commits to `devel` to pick up the fix, and this should be good to go. I'm :+1: on this being merged as soon as it passes CI checks

PR #1481, opened 2025-10-31
Fix contact groups

dchiquito:
## Related Issue

Fixes #1182 
## New Behavior

[Netbox 4.3.0](https://github.com/netbox-community/netbox/releases/tag/v4.3.0) changed the `Contact.contact_group` column to `Contact.contact_groups`, which subtly broke the `netbox_contact` module.

* In the `netbox_contact` module, add the `contact_groups` field alongside the existing `contact_group`.
  * 
* Add integration tests for adding contacts to contact groups.
* Add integration tests for contact groups in general.

## Contrast to Current Behavior

To add a contact to a contact group, you must now use one of two fields: `contact_group` or `contact_groups`, depending on your Netbox version. An informative error will be thrown if you use the wrong one.

## Discussion: Benefits and Drawbacks

Anyone who used the workarounds described in #1182 will probably need to update their playbooks.
The new API is much nicer IMHO.

## Changes to the Documentation

No changes besides the updated auto module docs. The `contact_group`/`contact_groups` fields document which Netbox versions are valid.

## Proposed Release Note Entry

Change `netbox_contacts.contact_group` to `contact_groups` (see #1182)

## Double Check

* [x] I have read the comments and followed the [CONTRIBUTING.md](https://github.com/netbox-community/ansible_modules/blob/devel/CONTRIBUTING.md).
* [x] I have explained my PR according to the information in the comments or in a linked issue.
* [x] My PR targets the `devel` branch.

PR #1488, opened 2025-11-13
Fix bad-return-value-key issue in sanity checks

sc68cal:
This changes "virtual machine" to "virtual_machine" in order to satisfy ansible-test's sanity checks, which has flagged this key as unusable with Jinja templating.


sc68cal:
This was fixed with a commit to #1481 so no need.

Issue #1404, opened 2025-04-08
[Bug]: netbox_tag module: slug not set as defined


jfunke4:
### Ansible NetBox Collection version

v3.20.0

### Ansible version

```bash
ansible [core 2.15.4]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /runner/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.16 (main, May 31 2023, 12:21:58) [GCC 8.5.0 20210514 (Red Hat 8.5.0-18)] (/usr/bin/python3.9)
  jinja version = 3.1.2
  libyaml = True
```

### NetBox version

v4.0.11

### Python version

3.9

### Steps to Reproduce

Create a tag with the netbox_tag module and define a name and a slug for this new tag.
When the slug contains two dashes (--) the netbox modules creates the tag with a slug, that only one dash (-).

Example:
Module: 
  netbox.netbox.netbox_tag:
    data:
      name: "customer123"
      slug: "customer--123"

Result in Netbox:
slug "customer-123"

### Expected Behavior

Creation of the tag as defined in the module.

### Observed Behavior

One dash is removed during the creation of the Tag

PR #1499, opened 2025-12-10
Fix regex pattern for character conversion

mabezi:
## Related Issue

#1404 

## New Behavior

Correct replacement of dashes and dots in slugs.

Incorrect replacement leads to errors when using slugs with multiple consecutive dashes.

This would make it possible to use tags like `customer--123` without errors.
There are also open source projects (e.g. [Yaook](https://gitlab.com/yaook/bare-metal/ansible-playbooks/-/blob/cc765a94fef39f66f1dde268300e7fbda23e1b3f/roles/netbox-config/defaults/main.yml#L93-107)) that frequently use multiple consecutive dashes in Netbox tags (`yaook--default-gateway`).

At this point, however, I would like to question why dashes are replaced in slugs at all, since dashes are valid characters in slugs.

not needed

> Fixed regex pattern for character conversion when using slug parameter in multiple modules

## Double Check

### DIFF
--- a/changelogs/fragments/1499-fix-regex-pattern.yml
+++ b/changelogs/fragments/1499-fix-regex-pattern.yml
@@ -1,4 +1,0 @@
----
-bugfixes:
-  - Fixed regex pattern for character conversion for slugs
-  - Removed useless character conversion when using slug parameter in multiple modules
--- a/changelogs/fragments/1499-rework-slugify.yml
+++ b/changelogs/fragments/1499-rework-slugify.yml
@@ -1,0 +1,7 @@
+---
+bugfixes:
+  - Adjust regex pattern for character conversion for slugs
+  - Only do slug generation in netbox_secret if needed
+  - Make slug generation optional by providing `slug` key
+breaking_changes:
+  - This changes the behavior for generating slugs. If you were relying on the previous behavior, you need to be aware that this behavior has changed.


plugins/module_utils/netbox_utils.py
```diff
@@ -1426,7 +1426,7 @@ def _to_slug(self, value):
             return value
         else:
             removed_chars = re.sub(r"[^\-\.\w\s]", "", value)
-            convert_chars = re.sub(r"[\-\.\s]+", "-", removed_chars)
+            convert_chars = re.sub(r"[\-\.\s]", "-", removed_chars)
```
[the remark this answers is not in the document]
sc68cal: Can I get an example of using the API where you try and create a slug that is invalid characters? Because my main concern is having an error message that users don't understand and they open bug tickets. As long as the error response from NetBox describes the issue properly, I'm all for removing our client side validation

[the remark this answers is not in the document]
mabezi: I tested it directly with the REST API instead of pynetbox; the effect should be the same.

Request:
```
curl -X 'POST' \
  'https://demo.netbox/api/extras/tags/' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -H 'X-CSRFTOKEN: <REMOVED>' \
  -d '{
  "name": "test",
  "slug": "test.123"
}'
```

Response with HTTP 400:
```
{
  "slug": [
    "Enter a valid “slug” consisting of Unicode letters, numbers, underscores, or hyphens."
  ]
}
```

mabezi:
I’ve now given this a lot of thought and considered how best to solve it.

Simply adjusting the regular expressions one-to-one isn't sufficient, because at the moment slugs are being replaced even when one is explicitly provided. Completely removing the function doesn’t make sense either, since generating slugs from the name is a desired feature.

This led to the following solution:
- I adapted the regular expressions used for **generating** slugs in NetBox.
- If a slug is provided as a parameter, the existing slug is preserved and not overwritten.

mabezi:
The unit tests do not agree with my solution yet. Well 🤨

mabezi:
Fixed & rebased to devel.

The CI errors that still occur should have nothing to do with my changes.

sc68cal:
CI tests are green on `devel` branch and other PRs don't have this error. I think unfortunately these changes are causing a regression and needs to be investigated.

mabezi:
I have found my mistake in thinking and developed a new plan!

From now on it will be like this:
- The regex is adapted to Netbox source -> **Important: In rare cases, this could possibly lead to deviations in how individual slugs are generated!** (example: `.Hello World.` is now `hello-world` instead of `-hello-world-` -> maybe we want to highlight this in changelog?)
- `netbox_secret` does not use character conversion for slug if there is a slug provided
- If a slug is specified as a parameter, no character conversion is performed. -> If someone relies on the `slug` parameter being automatically processed with `_to_slug`, this will now lead to a different result.

A very simplified example:
```yaml
tasks:
    - name: Create IP address within NetBox with only required information
      netbox.netbox.netbox_ip_address:
          netbox_url: http://netbox.local
          netbox_token: thisIsMyToken
          data:
              address: 192.168.1.10
              tags:
                  - "This Tag will be replaced with proper slug"
          state: present

    - name: Create IP address within NetBox with only required information
      netbox.netbox.netbox_ip_address:
          netbox_url: http://netbox.local
          netbox_token: thisIsMyToken
          data:
              address: 192.168.1.10
              tags:
                  - slug: "this-tag-wont-be-replaced-with-proper-slug"
          state: present
```

This should provide maximum compatibility.

sc68cal:
Great job, I know that this has turned into a pretty difficult thing to manage, since you have to deal with a lot of corner cases. I think it's a good idea to try and document this new behavior as much as possible as a `changelog` fragment.

For the most part when I use NetBox, the slugs were generated either by NetBox via the web application and the slugs are just used in my Ansible playbooks to properly reference things after the fact. My hope is that most people are fetching the slug values and just using them, and aren't really depending on the specific corner case behaviors that you have described. That should hopefully take some of the pressure off.

 I could be wrong, but that's at least my hope.

I'll do a deeper dive on your work at a later time so I'm fully up to speed with you, but great job

mabezi:
Is it already clear whether the patch is included with the next version? :)

Thanks in advance.

changelogs/fragments/1499-fix-regex-pattern.yml
```diff
@@ -0,0 +1,4 @@
+---
+bugfixes:
+  - Fixed regex pattern for character conversion for slugs
+  - Removed useless character conversion when using slug parameter in multiple modules
```
sc68cal: I think this needs a breaking change section.

```suggestion
  - Removed useless character conversion when using slug parameter in multiple modules
breaking_change:
  - This removes the old behavior for generating slugs. If you were relying on the previous behavior (i.e. replacing multiple dashes with a single dash, for example), you need to be aware that this behavior has been changed.
```

tests/unit/module_utils/test_data/netbox_utils/normalize_data.json
```diff
@@ -215,14 +215,6 @@
             "prefix_role": "test-role-1"
         }
     },
-    {
-        "before": {
-            "slug": "Test to_slug"
-        },
-        "after": {
-            "slug": "test-to_slug"
-        }
```
sc68cal, line 224: We still need this test though, right? It may just need to be adjusted? I ask because of your [example](https://github.com/netbox-community/ansible_modules/pull/1499#issuecomment-3762383169) where you have a tag `"This Tag will be replaced with proper slug"`  -it would be worth ensuring that this works correctly instead of removing the test.

changelogs/fragments/1499-fix-regex-pattern.yml
mabezi: Thank you for your response.
I omitted the example because it is not universally true, but formulated the changes much more precisely.

tests/unit/module_utils/test_data/netbox_utils/normalize_data.json
mabezi: I added a test scenario to see if the slug remains unedited.

I'm currently lacking ideas on how to test whether a slug is generated automatically, without rewriting much more tests out of this scope. (Correct generation is already tested separately with data from `slug.json`.)
With this PR a slug is only generated if `slug` is not set, so `"before": { "slug": "test test" }` would not work.

changelogs/fragments/1499-rework-slugify.yml
```diff
@@ -0,0 +1,7 @@
+---
+bugfixes:
+  - Adjust regex pattern for character conversion for slugs
+  - Only do slug generation in netbox_secret if needed
+  - Make slug generation optional by providing `slug` key
+breaking_changes:
+  - This changes the behavior for generating slugs. If you were relying on the previous behavior, you need to be aware that this behavior may have changed.
```
sc68cal:
```suggestion
  - This changes the behavior for generating slugs. If you were relying on the previous behavior, you need to be aware that this behavior has changed.
```

tests/unit/module_utils/test_data/netbox_utils/normalize_data.json
sc68cal: Ok, understood. Thank you for clarifying

tests/unit/module_utils/test_data/netbox_utils/normalize_data.json
```diff
@@ -217,10 +217,10 @@
     },
     {
         "before": {
-            "slug": "Test to_slug"
+            "slug": "already--valid_slug"
```
sc68cal, line 220: Perfect, thanks

sc68cal (approved):

Issue #1489, opened 2025-11-14
[Bug]: AWX can't install collection - Mismatch artifact hash with downloaded file


chupkes:
### Ansible NetBox Collection version

v3.21.0

### Ansible version

```bash
ansible [core 2.16.3]
  config file = None
  configured module search path = ['/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True

```

### NetBox version

v4.4.6

### Python version

3.11

### Steps to Reproduce

Try syncing a project in AWX with netbox.netbox in the requirements.yml. 

### Expected Behavior

The collection to install.

### Observed Behavior

Since yesterday, my project in AWX won't sync, because it errors on installing requirements. The cause seems to be the netbox.netbox collection. Worked fine before yesterday. Still works fine when you set the version to 3.20.0.

    "Downloading https://galaxy.ansible.com/api/v3/plugin/ansible/content/published/collections/artifacts/netbox-netbox-3.21.0.tar.gz to /var/lib/awx/projects/.__awx_cache/_15__bitbucket_ansible_playbooks_networking/stage/tmp/ansible-local-1696huk3za5_/tmp9dsa1f07/netbox-netbox-3.21.0-whb70f5_"
  ],
  "stderr_lines": [
    "ERROR! Failed to download collection tar from 'server0' due to the following unforeseen error: Mismatch artifact hash with downloaded file"
  ],

jplitza:
Apparently, @sc68cal [manually triggered the release](https://github.com/netbox-community/ansible_modules/actions/runs/19342712382) for commit b05a2d296dda6294f578547616b38e72292aa0f4 (which is based on the one I linked above).

So maybe he can trigger the release again for the original commit 90a506f43fd94e3f782f5b49858fe36ec9bafbca of the tag 3.21.0 to fix the situation?

Issue #1465, opened 2025-09-14
[Bug]: Reuploaded 3.21.0 which should be 3.22.0 ??


wdoekes:
### Ansible NetBox Collection version

v3.21.0

### Ansible version

I ran into this bug:
https://github.com/netbox-community/ansible_modules/issues/1426

And when I tried to fix it, I discovered that it was fixed in the currently installed version v3.21.0, **except I had to force-install it.**

**Turns out there are two v3.21.0.** See this:

```
ansible_modules$ git log -1 v3.21.0 
commit 90a506f43fd94e3f782f5b49858fe36ec9bafbca (tag: v3.21.0, origin/master)
Merge: de3ef6c f013f45
Author: Martin Rødvand <martin@rodvand.net>
Date:   Tue Mar 4 23:33:51 2025 +0100

    Merge pull request #1389 from netbox-community/devel
    
    Release 3.21
```
But also see this:

<img width="706" height="323" alt="Image" src="https://github.com/user-attachments/assets/a51df4c9-2aa0-4b13-9f89-17009e525c52" />

That's *also* netbox v3.21.0, updated **5 days ago**.

See this diff:

```diff
--- ./3.21.0-old/collections/ansible_collections/netbox/netbox/MANIFEST.json	2025-06-12 17:01:32.000000000 +0200
+++ ./3.21.0-new/collections/ansible_collections/netbox/netbox/MANIFEST.json	2025-09-14 11:06:30.000000000 +0200
@@ -1,34 +1,34 @@
 {
  "collection_info": {
   "namespace": "netbox",
   "name": "netbox",
   "version": "3.21.0",
   "authors": [
    "Mikhail Yohman <mikhail.yohman@gmail.com>",
    "Martin R\u00f8dvand <martin@rodvand.net>"
   ],
   "readme": "README.md",
   "tags": [
    "dcim",
    "system",
    "ipam",
    "networking"
   ],
   "description": "This is a collection of NetBox Ansible modules",
   "license": [],
   "license_file": "LICENSE",
   "dependencies": {},
   "repository": "https://github.com/netbox-community/ansible_modules",
   "documentation": "https://netbox-ansible-collection.readthedocs.io/en/latest/",
   "homepage": "https://github.com/netbox-community/ansible_modules",
   "issues": "https://github.com/netbox-community/ansible_modules/issues"
  },
  "file_manifest_file": {
   "name": "FILES.json",
   "ftype": "file",
   "chksum_type": "sha256",
[...]

PR #1500, opened 2025-12-15
ci: Fix release workflow running too often

jplitza:
## Related Issue

#1465
#1489 

## New Behavior

Don't run the release action of the CI pipeline for every new branch. Instead, only run it for tags that start with "v"

The action runs on every branch creation, as visible [here](https://github.com/netbox-community/ansible_modules/actions/workflows/release.yml)

* No more accidential releases to Ansible Galaxy

n/a

n/a

## Double Check

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 ---
 name: Deploy collection to Ansible Galaxy
 on:
-  create:
+  push:
     tags:
       - "v*"
 jobs:


sc68cal (approved):

jplitza:
> Maybe the regex needs to be changed to `^v*` since `v*` would match the `vel` in `devel`?

As I understand [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs), it isn't a regular expression, but a glob. As such, v* should be correct. Also, it never matches on branches, only on tags.

Issue #1444, opened 2025-07-18
[Bug]: nb_inventory func extract_dns_name() doesn't handle unassigned interface ip address


jbroillet:
### Ansible NetBox Collection version

v3.21.0

### Ansible version

```bash
ansible --version
ansible [core 2.18.6]
  config file = /home/user/netdevops/semaphore/network-automation/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/netdevops/semaphore/network-automation/.venv/lib64/python3.11/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/netdevops/semaphore/network-automation/.venv/bin/ansible
  python version = 3.11.11 (main, Feb 10 2025, 00:00:00) [GCC 11.5.0 20240719 (Red Hat 11.5.0-5)] (/home/user/netdevops/semaphore/network-automation/.venv/bin/python3.11)
  jinja version = 3.1.6
  libyaml = True
```

### NetBox version

v4.3.2

### Python version

3.11

### Steps to Reproduce

1. In NetBox, create a device with the following setup:
   - A private IP address (e.g., `192.168.1.1`) assigned to an interface (e.g., `GigabitEthernet0/0`).
   - A public/NAT IP address (e.g., `203.0.113.1`), set NAT INSIDE field to the created assigned interface (e.g., `192.168.1.1`) and set the NAT IP address (e.g., `203.0.113.1`) as the **primary IP** of the device, but **not** assigned to any interface.

2. Observe the behavior:
   - The function fails to resolve the DNS name correctly and produces an error when the primary IP is not associated with an interface.
   - Inventory plugin completely fail and return any hosts

### Expected Behavior

The `extract_dns_name()` function should be able to extract a valid DNS name even if the device's primary IP address is not assigned to any interface.

In this scenario, since the primary IP (e.g., the NAT outside address) is intentionally not associated with an interface, the function should still return a usable DNS name based on the primary IP rather than failing.

### Observed Behavior

I have observed that the `extract_dns_name()` function does not handle the case where the primary IP is not assigned to an interface. The relevant portion of the function is:

```python
def extract_dns_name(self, host):
    # No primary IP assigned
    if not host.get("primary_ip"):
        return None

    before_netbox_v29 = bool(self.ipaddresses_lookup)
    if before_netbox_v29:
[...]

PR #1514, opened 2026-01-21
Fix faulty IP address DNS check

dylan-stella:
## Related Issue

#1444 

## New Behavior

Check for no IP address being available

Current behavior breaks inventory sync if the primary IP is not on one of the interfaces when checking for DNS. If you're not using the DNS feature, then this is the same as having None.

This breaks our inventory sync periodically when there is dirty data left behind in our Netbox. Since we're not using the DNS feature and we don't need to check.

However, this will hide the issue where the primary IP is not on any interface.

None

Fix sync error when primary IP address not on interfaces

## Double Check

### DIFF
--- a/changelogs/fragments/fix-ip-address-dns-check.yml
+++ b/changelogs/fragments/fix-ip-address-dns-check.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - nb_inventory - Fix AttributeError when `ip_address` is None during DNS name lookup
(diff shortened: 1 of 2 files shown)


sc68cal:
Hi, thanks for the contribution. If you could please create a `changelog` fragment, that would get CI passing. These are required so that we can generate release notes (and give you credit) when we create a new release

sc68cal:
The linting failure, I think it's because `black` pushed a new version with some new rules. You will need to rebase, once it is fixed

sc68cal (approved):

PR #1522, opened 2026-02-10
nb_inventory - fix cache key uniqueness across sources

### LINKED ISSUE #1411: Validate compatibility with ansible-core 2.19
Copied from ansible-community/ansible-build-data#538

 ansible-core has implemented an extensive rewrite of templating and error handling as part of the [data tagging feature ](https://github.com/ansible/ansible/pull/84621) coming in core-2.19. You can find information about this change in that PR description, as well as in the [core 2.19 porting guide](https://docs.ansible.com/ansible-core/devel/porting_guides/porting_guide_core_2.19.html). 
 
 Please read  [background and testing requests,](https://forum.ansible.com/t/data-tagging-preview-and-testing/40759) which details where you can file potential bugs.

Collections that are part of the Ansible community package MUST test against `ansible-core 2.19` to ensure they continue to work with the updated core. Collections which are not compatible with `ansible-core 2.19`  will be removed from Ansible 12.  You can choose to test against the `stable-2.19` branch, the `devel` branch, the `milestone` branch, or the pre-release/released versions of `ansible-core 2.19` available on [Pypi](https://pypi.org/project/ansible-core/#history).

This should be an ongoing verification/check until the release of core 2.19 to ensure compatibility with any bugfixes or changes that happen until then.

You can find advice from fellow maintainers at [Making a collection compatible with data tagging](https://forum.ansible.com/t/making-a-collection-compatible-with-data-tagging/41717), and we welcome your additions to that forum topic.
 
 **Note:** You are not required to implement data tagging as part of this request, though we do recommend you consider where it can improve the quality of your collection. This issue is focused on ensuring you have verified your collection works with `ansible-core 2.19` in order to remain included in the Ansible 12 community package.
 
 We ask that you resolve any problems and link those PRs to this issue. If you do not find any problems with your integration/unit testing, please provide that comment and a pointer to your test results before closing this issue as done.
 
 You can reach out on the [Ansible forum](https://forum.ansible.com/t/data-tagging-preview-and-testing/40759) for more help or discussion.

s-hertel:
## Related Issue

~Possibly fixes https://github.com/netbox-community/ansible_modules/issues/1411#issuecomment-3834941933 / https://github.com/ansible/ansible/issues/86292.~

~I suspect a large inventory is needed to reproduce the performance issue. I was not able to reproduce by using a small inventory provided by the netbox free trial. The issue reported against ansible/ansible outputs \~30x as many lines as my attempt to reproduce.~

No related issues.

## New Behavior

Pass `get_cache_key` the inventory source, and read/set the backing cache once per source. This aligns with the developer guide and intended use of the method: https://docs.ansible.com/projects/ansible/latest/dev_guide/developing_inventory.html#inventory-cache.

This ~reduces the number of times the raw data is converted to/from the internal cache format, and~ fixes cache key uniqueness across sources.

~Every HTTP request is converted to/from the internal cache format.~

Multiple `netbox.netbox.nb_inventory` sources can read from/write to the same cache keys, which is precisely what the `get_cache_key` method is intended to prevent.

The change is backwards compatible.

N/A

N/A

## Double Check

### DIFF
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -2169,9 +2169,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         cache_key = self.get_cache_key(path)
 
         global UPDATE_CACHE
-        UPDATE_CACHE = not self.use_cache and self.get_option('cache')
+        UPDATE_CACHE = not self.use_cache and self.get_option("cache")
 
-        if self.use_cache and self.get_option('cache'):
+        if self.use_cache and self.get_option("cache"):
             try:
                 HTTP_RESPONSE.update(self._cache[cache_key])
             except KeyError:
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -422,13 +422,18 @@ except ImportError as imp_exc:
 else:
     PYTZ_IMPORT_ERROR = None
 
+# Temporary cache
+HTTP_RESPONSE = {}
+
+# A global indicating whether the backing cache should be updated
+UPDATE_CACHE = False
+
 
 class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
     NAME = "netbox.netbox.nb_inventory"
 
     def _fetch_information(self, url):
         results = None
-        cache_key = self.get_cache_key(url)
 
         # get the user's cache option to see if we should save the cache if it is changing
         user_cache_setting = self.get_option("cache")
@@ -439,12 +444,14 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         # attempt to read the cache if inventory isn't being refreshed and the user has caching enabled
         if attempt_to_read_cache:
             try:
-                results = self._cache[cache_key]
+                results = HTTP_RESPONSE[url]
                 need_to_fetch = False
             except KeyError:
-                # occurs if the cache_key is not in the cache or if the cache_key expired
+                # occurs if the URL is not in the local cache
                 # we need to fetch the URL now
                 need_to_fetch = True
+                global UPDATE_CACHE
+                UPDATE_CACHE = self.get_option("cache")
         else:
             # not reading from cache so do fetch
             need_to_fetch = True
@@ -491,9 +498,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             except ValueError:
                 raise AnsibleError("Incorrect JSON payload: %s" % raw_data)
 
-            # put result in cache if enabled
-            if user_cache_setting:
-                self._cache[cache_key] = results
+            # put result in local cache if enabled
+            if UPDATE_CACHE:
+                HTTP_RESPONSE[url] = results
 
         return results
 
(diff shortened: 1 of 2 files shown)


sc68cal:
Can you discuss this change in a bit more detail? I understand that you are trying to reduce the number of HTTP requests that are sent, but my concern is that it's using a global variable. Can you discuss the global variable usage as opposed to having the `InventoryModule` class contain the dictionary that you are using as the cache, and then getters/setters?

I ask this because the only uses I see of it are within `InventoryModule._fetch_information` and `InventoryModule.parse` which means that it would be very simple to have it as an internal field of the class.

Thanks.

s-hertel:
It can certainly use an internal field instead of a global. To clarify, this has nothing to do with reducing the number of HTTP requests.

When parsing multiple inventory sources with this plugin (for example  `ansible-inventory -i netbox1.yml -i netbox2.yml --list`), nothing prevents those sources from using the same cached data. A user can work around it by configuring different backends or ensuring no sources use the same `api_endpoint`, but normally this is handled by default, by passing the inventory source to `self.get_cache_key`. This way if netbox1.yml has access to some objects, and netbox2.yml has access to other objects, they retrieve their own data instead of each other's. (Note: I'm assuming netbox object-level permissions are similar to other cloud services.)

I hoped reducing the number of times the cache needs to be deserialized would improve the performance, however, it looks like that is [not the case](https://github.com/ansible/ansible/issues/86292#issuecomment-3883077343). In light of that, I've simplified this change.

sc68cal:
Awesome. Thank you for the detailed write-up, I am now in sync with your thinking. This change looks good to me

sc68cal (dismissed):

plugins/inventory/nb_inventory.py
```diff
@@ -428,7 +428,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
     def _fetch_information(self, url):
         results = None
-        cache_key = self.get_cache_key(url)
+        cache_key = self.cache_key_base + self.get_cache_key(url)
```
sc68cal, line 431: Can you explain this a bit more? Because it still isn't quite clear to me. You set `cache_key_base` down in `parse` with the results of `self.get_cache_key(path)` and then just add it together with a second call to `self.get_cache_key(url)` but it's not clear to me why you are doing this.

I would like to get a little more context on why you are doing this the way you are doing it.

I did take a look at some other Ansible plugins that call `self.get_cache_key` and none of them appear to do the behavior that you are introducing, so I'd like to understand it a bit more.

I also read the link that you provided from the Ansible plugin development guide and they too, do not appear to call out this issue you describe.

> When parsing multiple inventory sources with this plugin (for example ansible-inventory -i netbox1.yml -i netbox2.yml --list), nothing prevents those sources from using the same cached data

In your example, are these two netbox inventory sources pointing at the same NetBox server, or different servers?

Thanks!

s-hertel: This ensures separate inventory sources use separate cache keys by adding a unique but predictable prefix created from the inventory source.

> I did take a look at some other Ansible plugins that call self.get_cache_key and none of them appear to do the behavior that you are introducing, so I'd like to understand it a bit more.

Most inventory plugins call `get_cache_key` per inventory source, not per URL. I've added the former, but haven't removed the latter (mainly because I'm not sure whether or not it's a significant issue, however, you might want to look at https://github.com/ansible/ansible/issues/86504 doing something very similar). Here's an example of how inventory plugins should use `self.get_cache_key`: https://github.com/ansible-collections/amazon.aws/blob/11.1.0/plugins/inventory/aws_ec2.py#L964 -> https://github.com/ansible-collections/amazon.aws/blob/11.1.0/plugins/plugin_utils/inventory.py#L193.

> In your example, are these two netbox inventory sources pointing at the same NetBox server, or different servers?

Both inventory sources need to call the same URL(s) to reproduce what I'm talking about. You can test by setting `token` to a valid value in the first file, and omit it from the second. If you compare with an inventory plugin like amazon.aws.aws_ec2, separate sources use their own authentication.

s-hertel:
I no longer have the means to test netbox, so I'll close this. Thank you for the reviews! I've updated the developer documentation https://docs.ansible.com/projects/ansible/devel/dev_guide/developing_inventory.html to advise inventory plugins not to use the `get_cache_key` helper multiple times since the intended use was not as clear as it should have been.

k0ste:
> I no longer have the means to test netbox, so I'll close this.

Can you write `tl;dr`? I search answer for "when Ansible 13 will work fast with NetBox again (like Ansible v9)?"

s-hertel:
> I search answer for "when Ansible 13 will work fast with NetBox again (like Ansible v9)?"

@k0ste I don't know the answer either. I cannot reproduce. Report it as a bug so Netbox maintainers investigate.

PR #1525, opened 2026-02-11
Enhancement: add option to trust templated strings

etrombly:
## Related Issue

#1411

## New Behavior

added config option named `trust_template_vars` to allow templated strings to render in ansible-core > 2.19.0.

templates are not currently rendered

It defaults to disabled, and I verify that the trust_as_template function is available, so it should be backwards compatible. It does modify the config in place, but I don't think that's an issue.

May want to add a note about the new behavior. I didn't add anything yet.

...

## Double Check


etrombly:
Bumping this for review. I believe a lot of people do use template vars, and ansible-core versions < 2.19 are all EOL now. So there is no currently supported way to use them.

theherox:
I'll also bump this for review.  I'm one of those people that @etrombly mentioned using templated vars extensively.

PR #1543, opened 2026-04-01
Add NetBox v4.5 CI support (INT-238)

ldrozdz93:
## Summary
- Add v4.5 (netbox-docker 4.0.2) to CI matrix
- Add docker-compose override with custom `super_user.py` that creates a v1 (legacy) API token for backward compatibility (v4.5 defaults to v2 tokens with `Bearer nbt_<KEY>.<TOKEN>` auth header)
- Add v4.5 integration test targets (91 task files, inventory, regression)

### Module updates for v4.5 breaking changes
- **front_port/front_port_template**: Make `rear_port` optional; version-conditional logic strips it for v4.5+ (replaced by PortMapping model) and validates presence for older versions
- **token**: Make `key` optional; v4.5+ v2 tokens use description-based lookup instead of key-based; runtime validation enforces key for pre-v4.5
- **user**: Strip `is_staff` from data for v4.5+ (field removed from User model)
- **inventory fixtures**: Regenerated all 10 JSON fixtures from live v4.5 instance

## Test plan
- [x] v4.5 module integration tests pass (ok=1099, failed=0, ignored=9)
- [x] v4.1-v4.4 module integration tests pass (backward compatibility verified)
- [x] Inventory fixtures regenerated and verified (10/10 pass)
- [x] Linting (black + yamllint) passes
- [x] Unit tests pass (Python 3.11, 3.12, 3.13)
- [x] All CI checks green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/module_utils/netbox_users.py
+++ b/plugins/module_utils/netbox_users.py
@@ -73,10 +73,6 @@ class NetboxUsersModule(NetboxModule):
                     self.module.fail_json(
                         msg="'key' is required for tokens on NetBox < v4.5"
                     )
-                if self.state == "absent" and not data.get("key"):
-                    self.module.fail_json(
-                        msg="'key' is required for state=absent on NetBox < v4.5"
-                    )
 
         # Used for msg output
         if data.get("username"):
@@ -97,6 +93,9 @@ class NetboxUsersModule(NetboxModule):
                 query_params["id"] = data["id"]
             elif data.get("description"):
                 query_params["description"] = data["description"]
+                # Scope by user so tokens with same description across users don't collide
+                if data.get("user"):
+                    query_params["user_id"] = data["user"]
             self.nb_object = self._nb_endpoint_get(nb_endpoint, query_params, name)
         else:
             object_query_params = self._build_query_params(
--- a/plugins/module_utils/netbox_users.py
+++ b/plugins/module_utils/netbox_users.py
@@ -90,10 +90,21 @@ class NetboxUsersModule(NetboxModule):
         else:
             name = "token"
 
-        object_query_params = self._build_query_params(
-            endpoint_name, data, user_query_params
-        )
-        self.nb_object = self._nb_endpoint_get(nb_endpoint, object_query_params, name)
+        # v4.5+ tokens: look up by description (not key) since v2 keys are auto-generated
+        if self.endpoint == NB_TOKENS and is_v45:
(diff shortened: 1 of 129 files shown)


plugins/module_utils/netbox_users.py
```diff
@@ -45,18 +45,66 @@ def run(self):
...
+                            "is required for idempotent operations (v2 tokens "
+                            "do not support key-based lookup)."
+                        )
+                    )
+            else:
+                if not data.get("key"):
+                    self.module.fail_json(
+                        msg="'key' is required for tokens on NetBox < v4.5"
+                    )
+                if self.state == "absent" and not data.get("key"):
```
unlisted, line 76: nit: unreachable if key is absent from data.

plugins/module_utils/netbox_users.py
```diff
@@ -45,18 +45,66 @@ def run(self):
...
+        else:
+            name = "token"
+
+        # v4.5+ tokens: look up by description (not key) since v2 keys are auto-generated
+        if self.endpoint == NB_TOKENS and is_v45:
+            query_params = {}
+            if data.get("id"):
+                query_params["id"] = data["id"]
+            elif data.get("description"):
+                query_params["description"] = data["description"]
```
unlisted, line 99: claude raises a good point, worth investigating.

> On v4.5+, the lookup queries by description only, without filtering by user. If two different users have tokens with
  the same description, this could match the wrong one (or trigger a "more than one result" error from
  _nb_endpoint_get). Copilot flagged this too but the user addition was reverted because it broke older versions. The
  runtime-override approach in commit 5 was the right fix for ALLOWED_QUERY_PARAMS, but the actual query construction
  still doesn't include user. Could you add user to the runtime query_params dict here (after resolving user to
  user_id)?

unlisted (approved):
Looking good, couple of comments but no blockers.

plugins/module_utils/netbox_users.py
ldrozdz93: Good catch — removed the redundant check.

plugins/module_utils/netbox_users.py
ldrozdz93: Applied — added `user_id` to the v4.5+ token lookup query when `description` is used, so tokens from different users with the same description don't collide.

PR #1555, opened 2026-04-25
Add primary_ip_interface option to nb_inventory

### LINKED ISSUE #1530: [Feature]: Retrieve primary_interface_name variable from netbox.netbox.nb_inventory plugin.
### NetBox version

v4.5.3

### Feature type

Change to existing Plugin

### Proposed functionality

The netbox.netbox.nb_inventory currently does not output the name of the interface of the primary ip.
interfaces: true, quickly becomes too much and decreases speed drastically.

Currently using collection version: v3.22.0

### Use case

Would be great to have the interface name of the primary interface returned so one can set this dynamically as inventory variables.

### External dependencies

_No response_

jorgenspange:
## Summary

- Adds a new `primary_ip_interface` boolean option to the `nb_inventory` plugin that exposes the interface name of the primary IP as a host var (`primary_ip_interface`)
- When only this option is set, a targeted fetch queries only the primary IP IDs already present on the fetched devices/VMs — rather than all IP addresses in NetBox — keeping overhead minimal (~0.6s in testing)
- When used alongside `interfaces: true`, `dns_name: true`, or `ansible_host_dns_name: true`, the existing full IP fetch is reused with no duplicate requests

Closes #1530

## Test plan

- [x] Unit tests updated and passing (`test_group_extractors`)
- [x] Changelog fragment added (`changelogs/fragments/1530-primary-ip-interface.yml`)
- [x] Enable `primary_ip_interface: true` in an inventory config and verify the `primary_ip_interface` host var is populated with the correct interface name
- [x] Verify hosts without a primary IP, or whose primary IP is not assigned to an interface, have no `primary_ip_interface` var set
- [x] Verify no regression when `primary_ip_interface: false` (default)

### DIFF
--- a/changelogs/fragments/1530-primary-ip-interface.yml
+++ b/changelogs/fragments/1530-primary-ip-interface.yml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-  - nb_inventory - Add `primary_ip_interface` option to expose the interface name of the primary IP as a host var without requiring `interfaces: true`
+  - nb_inventory - Add ``primary_ip_interface`` option to expose the interface name of the primary IP as a host var without requiring ``interfaces`` to be enabled
--- a/changelogs/fragments/1530-primary-ip-interface.yml
+++ b/changelogs/fragments/1530-primary-ip-interface.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - nb_inventory - Add `primary_ip_interface` option to expose the interface name of the primary IP as a host var without requiring `interfaces: true`
--- a/plugins/inventory/nb_inventory.py
+++ b/plugins/inventory/nb_inventory.py
@@ -1466,7 +1466,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             if ipaddress.get("assigned_object_type") == "virtualization.vminterface":
                 self.vm_ipaddresses_lookup[ip_id] = {"interface_name": interface_name}
             else:
-                self.device_ipaddresses_lookup[ip_id] = {"interface_name": interface_name}
+                self.device_ipaddresses_lookup[ip_id] = {
+                    "interface_name": interface_name
+                }
 
     def refresh_interfaces(self):
         url_device_interfaces = self.api_endpoint + "/api/dcim/interfaces/?limit=0"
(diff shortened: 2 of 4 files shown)


jorgenspange:
@sc68cal Would you be able to take a look at this one, when you have got the time?

sc68cal:
This seems like something that can already be done, via an `nb_lookup` query, for those that require it. Can you discuss why this should be used instead of a `nb_lookup` query?

jorgenspange:
I have to admit that I am not familiar with the nb_lookup query, where would I find the documentation for this?

jorgenspange:
I think I got it now, looking into the nb_lookup plugin. Using the lookup_plugin would require me to run this task for all my hosts while having it in the inventory plugin makes primary_ip_interface available as a native host var before any play runs, just like primary_ip4 already. This is also consistent with how dns_name is already handled in the inventory plugin.

jorgenspange:
> I think I got it now, looking into the nb_lookup plugin. Using the lookup_plugin would require me to run this task for all my hosts while having it in the inventory plugin makes primary_ip_interface available as a native host var before any play runs, just like primary_ip4 already. This is also consistent with how dns_name is already handled in the inventory plugin.

Any thoughts on this @sc68cal?

sc68cal:
My concern is that we already have a couple of different ways that people want to connect to devices, by using different settings in the netbox inventory plugin. I don't think we should continue to add more and more different ways. If you have some preferred way to connect to your device based on netbox data that is not `primary_ip4` or `dns_name`, you should use the `nb_lookup` plugin with `add_host` or set host_vars from the inventory data. I don't think we should keep growing the complexity in `nb_inventory` for each different way that people want to connect to devices that deviates from `primary_ip4`

jorgenspange:
@sc68cal This is not used to connect to the device really, what I am mostly interested in is to retreive the primary interface name, so that I can populate this in different variables I have in my configuration.

sc68cal:
> what I am mostly interested in is to retreive the primary interface name, so that I can populate this in different variables I have in my configuration.

That's a usecase that the `nb_lookup` plugin is meant for.

jorgenspange:
@sc68cal thank you so much.

Seems like I was able to solve it like this with the lookup plugin:
```yaml
ntp_source_interface: >-
  {{ ((lookup('netbox.netbox.nb_lookup', 'ip-addresses',
      api_endpoint='https://netbox.example.com',
      api_filter='address=' + ansible_host,
      validate_certs=False,
      raw_data=True) | default({})).assigned_object | default({})).name | default(none) }}
```

PR #1571, opened 2026-06-10
fix: compare unordered m2m list fields as sets in the generic diff engine (INT-411)

ldrozdz93:
## Summary

Fixes idempotency failures caused by the update-comparison engine treating unordered many-to-many (m2m) list fields as **ordered** lists. NetBox returns m2m fields in a non-deterministic order, so a task whose list contents already match NetBox state is reported as `changed` on every run, and `--check` mode reports phantom changes.

The engine previously special-cased only `tags` for set comparison; every other list field was compared element-wise. A per-module fix shipped for `netbox_permission` (`LIST_AS_SET_KEYS` in `netbox_users.py`), but the bug class stayed live for fields flowing through the generic engine — most visibly `netbox_tag` with `object_types`, which uses the generic engine and exhibits the exact #1486 behaviour today.

## What changed

- **`netbox_utils.py`** — a single `LIST_AS_SET_KEYS` set now drives set comparison in the generic `_update_netbox_object`, replacing the `tags`-only special case. The conversion only applies when both the existing object and the incoming data carry the key, and falls back to ordered comparison when elements aren't hashable.
- The set covers: `tags`, `object_types`, `tagged_vlans`, the `netbox_config_context` assignment scopes (`regions`, `sites`, `site_groups`, `locations`, `device_types`, `roles`, `platforms`, `cluster_types`, `cluster_groups`, `clusters`, `tenant_groups`, `tenants`), the route-target lists (`import_targets`, `export_targets`), and the user lists (`permissions`, `groups`, `actions`).
- **`netbox_users.py`** — imports the shared `LIST_AS_SET_KEYS` instead of redefining its own subset, so there is one source of truth.
- **Unit tests** — reordering an unordered list field reports no change; a genuine content difference is still detected; non-hashable list elements do not raise.
- **Integration test** (`v4.5/netbox_tag.yml`) — create a tag with `object_types`, re-run with the same values in a different order, assert `not changed`.
- **Changelog fragment.**

## Resolves

- GitHub #1486 (`netbox_permission.object_types` reordering) — verify & close.
- The latent, previously unreported `netbox_tag.object_types` case.

Linear: INT-411

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -758,7 +758,6 @@ LIST_AS_SET_KEYS = set(
         "regions",
         "site_groups",
         "sites",
-        "locations",
         "device_types",
         "roles",
         "platforms",
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -743,8 +743,10 @@ SCOPE_TO_ENDPOINT = {
 # order is wrongly reported as a change (breaking idempotency and --check mode).
 # Adding a field name here is safe even if it never appears on a given endpoint:
 # the comparison only converts a key when both the existing object and the
-# incoming data carry it, and non-hashable values fall back to ordered
-# comparison. See INT-411 and GitHub issue #1486.
+# incoming data carry it. Every listed field serializes (via pynetbox
+# Record.serialize) to hashable IDs or content-type/action strings; do not add a
+# field that serializes to dicts (e.g. cable terminations) — set() will raise,
+# on purpose. See INT-411 and GitHub issue #1486.
 LIST_AS_SET_KEYS = set(
     [
         "tags",
@@ -1579,15 +1581,8 @@ class NetboxModule(object):
         # reordering of identical values is not reported as a change.
         for key in LIST_AS_SET_KEYS:
             if serialized_nb_obj.get(key) and data.get(key):
-                try:
-                    before_set = set(serialized_nb_obj[key])
-                    after_set = set(data[key])
-                except TypeError:
-                    # Elements aren't hashable (e.g. a list of dicts); keep the
-                    # ordered comparison rather than failing.
-                    continue
-                serialized_nb_obj[key] = before_set
-                updated_obj[key] = after_set
+                serialized_nb_obj[key] = set(serialized_nb_obj[key])
+                updated_obj[key] = set(data[key])
 
         # Ensure idempotency for site on older netbox versions
         version_pre_30 = self._version_check_greater("3.0", self.api_version)
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -737,6 +737,42 @@ SCOPE_TO_ENDPOINT = {
     "virtualization.clustergroup": "cluster_groups",
 }
 
+# Many-to-many list fields that NetBox returns in a non-deterministic order.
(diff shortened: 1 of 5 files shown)


dhoffend:
I had similiar ideas and discussions in PR #1487 
But with your integration tests it is even better 👍🏻

dhoffend:
When fixing the ansible collection. It would make sense to sync the set fields with the ‘pynetbox’ modules as well

netbox-community/pynetbox#711

ldrozdz93:
> When fixing the ansible collection. It would make sense to sync the set fields with the ‘pynetbox’ modules as well
https://github.com/netbox-community/pynetbox/issues/711

This PR will fix all of the 19 keys susceptible for the ordering problem, not just `tags` and `tagged_vlans`.

unlisted (commented):
[claude]

One note on the unhashable fallback — inline.

plugins/module_utils/netbox_utils.py
```diff
@@ -1540,9 +1575,19 @@ def _update_netbox_object(self, data):
...
-            serialized_nb_obj["tags"] = set(serialized_nb_obj["tags"])
-            updated_obj["tags"] = set(data["tags"])
+        # Compare unordered many-to-many list fields as sets so a pure
+        # reordering of identical values is not reported as a change.
+        for key in LIST_AS_SET_KEYS:
+            if serialized_nb_obj.get(key) and data.get(key):
+                try:
+                    before_set = set(serialized_nb_obj[key])
+                    after_set = set(data[key])
+                except TypeError:
```
unlisted: [claude]
The `except TypeError: continue` is correct as a guard, but it has a quieter consequence: for any field where pynetbox `.serialize()` yields *unhashable* elements (a list of dicts), this silently skips set conversion and a reorder still reports `changed` — i.e. the fix becomes a no-op for that field while the changelog implies it's fixed. The tests only prove `tags`/`object_types`/`tagged_vlans` with synthetic int/string lists plus the fallback, and the only field exercised against a real endpoint is `object_types` (the `netbox_tag` integration test). In practice `.serialize()` returns IDs/strings (hashable) for the route-target and config-context-scope fields too, so this likely works — but it's unverified for everything except `object_types`. Worth either confirming the serialized shape for at least one route-target / config-context case, or noting the "hashable only" limitation explicitly so a future reader doesn't assume blanket coverage.

unlisted (dismissed):
looking good! claude anchored comment may be worth addressing.

plugins/module_utils/netbox_utils.py
ldrozdz93: [claude]
Addressed by removing the silent fallback entirely (commit `4f4acb0`).

I confirmed the serialized shape you flagged: pynetbox `Record.serialize()` collapses FK-style m2m list items to their integer `id` (`pynetbox/core/response.py` L684), and only id-less nested mappings (cable terminations, `rear_ports`, `link_peers`, `connected_endpoints`) become dicts — none of which are in `LIST_AS_SET_KEYS`. So the `except TypeError` branch never fired for any current field.

Rather than keep a silent no-op path (exactly what could hide a future misuse while the changelog claims a fix), the loop now does a plain `set(...)`: if someone ever adds a dict-serializing field, it raises loudly at that spot instead of silently degrading. A code comment states the invariant (listed field ⇒ hashable serialized elements). This also realigns the generic engine with `netbox_users.__update_netbox_object__`, which already used a bare `set(...)`.

ldrozdz93:
I bypassed to merge, because the previous approval should still be valid.

PR #1574, opened 2026-06-10
fix(netbox_device_interface): persist legacy mac_address on NetBox v4.2+

### LINKED ISSUE #1502: [Bug]: netbox.netbox.netbox_device_interface return changed for old mac_address but doesn't change it

v.3.21.0

### Ansible version

ansible [core 2.20.1]
  config file = /Users/aivanov/Git/asset-automations/ansible.cfg
  configured module search path = ['/Users/aivanov/Git/asset-automations/modules']
  ansible python module location = /Users/aivanov/Git/asset-automations/venv/lib/python3.13/site-packages/ansible
  ansible collection location = /Users/aivanov/Git/asset-automations/collections
  executable location = /Users/aivanov/Git/asset-automations/venv/bin/ansible
  python version = 3.13.7 (v3.13.7:bcee1c32211, Aug 14 2025, 19:10:51) [Clang 16.0.0 (clang-1600.0.26.6)] (/Users/aivanov/Git/asset-automations/venv/bin/python3.13)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)

### NetBox version

v4.4.8

### Python version

3.11

### Steps to Reproduce

1. Run task
```
- name: Updating management interface items in Netbox
  environment:
    REQUESTS_CA_BUNDLE: "{{ cert_path }}"
    SSL_CERT_FILE: "{{ cert_path }}"
  delegate_to: localhost
  netbox.netbox.netbox_device_interface:
    validate_certs: "{{ pdu_panduit_netbox_validate_certs }}"
    netbox_url: "{{ pdu_panduit_netbox_host }}"
    netbox_token: "{{ pdu_panduit_netbox_token }}"
    data:
      device: "{{ inventory_hostname }}"
      name: "management"
      mac_address: "{{ pdu_panduit_facts.main.mac_addr | replace('-', ':') | lower }}"
      type: "100base-tx"
      mgmt_only: true
    state: present
```

2. Observe output
```
TASK [pdu_panduit : Updating management interface items in Netbox] **********************************************************************************************************************************************************************
--- before
+++ after
@@ -1,3 +1,3 @@
 {
-    "mac_address": null
+    "mac_address": "00:0F:9C:XX:XX:XX"
 }

changed: [pdu-abcd -> localhost]
```

3. See no changes in Netbox. Because it expects to create the MAC address via another module.

### Expected Behavior

It must return an error indicating that the MAC address should be created first, or force-creating it under the hood to maintain backward compatibility. 

### Observed Behavior

The task shows changes as they were made, but it is not true.

ldrozdz93:
## Problem

On NetBox 4.2+, setting the legacy `mac_address` on `netbox_device_interface` reports `changed` on every run — including check mode — but stores nothing. NetBox 4.2 moved MACs to a standalone `MACAddress` model and made the interface field read-only, so the API silently drops the value and the module's local diff reappears forever. Users believe the MAC is in NetBox when it isn't.

## Solution

On NetBox 4.2+ the module now persists the value: it finds or creates a `MACAddress` assigned to the interface and sets it as the interface's `primary_mac_address`. Existing playbooks keep working, the task is idempotent, check mode reports a change only when one would really be made, and an explicitly supplied `primary_mac_address` always wins over the legacy field. NetBox < 4.2, where the field is still writable, is unchanged.

## Testing

Verified manually against a live NetBox 4.5.0: the first run creates exactly one assigned `MACAddress` and sets it primary; re-running the identical task reports no change and creates no duplicate; check mode writes nothing. Exercised both on an existing interface and when creating an interface with `mac_address` in a single task.

New unit tests cover the create-and-assign, reuse, already-primary, and check-mode paths. Integration plays for v4.2–v4.5 assert the MAC becomes primary, the second run is `ok`, and check mode reports no phantom change.

Fixes #1502

### DIFF
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -837,6 +837,85 @@ class NetboxModule(object):
 
         return t_greater > t_lesser if not greater_or_equal else t_greater >= t_lesser
 
+    def _set_primary_mac_address(
+        self, endpoint_name, name, mac_address, assigned_object_type
+    ):
+        """Store a legacy ``mac_address`` as the interface's primary MAC.
+
+        Find or create a ``MACAddress`` assigned to the interface and set it as
+        ``primary_mac_address``. The interface's read-only ``mac_address`` mirrors its
+        primary MAC, so a re-run that already has the right primary makes no change.
+        Called after the interface is ensured, so ``self.nb_object`` is the interface.
+        """
+        mac_address = mac_address.upper()
+        # After an update with no field changes, _ensure_object_exists leaves
+        # self.nb_object as the serialized dict rather than a pynetbox Record.
+        nb_object = self.nb_object
+        serialized = nb_object if isinstance(nb_object, dict) else nb_object.serialize()
+
+        # The interface's read-only `mac_address` already reflects its primary MAC.
+        if serialized.get("mac_address") == mac_address:
+            return
+
+        if self.check_mode:
+            self._mark_changed_by_mac(endpoint_name, name)
+            return
+
+        interface_id = serialized["id"]
+        mac = self._find_or_create_mac_address(
+            mac_address, assigned_object_type, interface_id
+        )
+        app = assigned_object_type.split(".")[0]
+        nb_endpoint = getattr(getattr(self.nb, app), self.endpoint)
+        if serialized.get("primary_mac_address") != mac.id:
+            nb_endpoint.get(interface_id).update({"primary_mac_address": mac.id})
+        # Re-fetch so the returned object's read-only `mac_address` reflects the primary.
+        self.nb_object = nb_endpoint.get(interface_id)
+
+        before = (self.result.get("diff") or {}).get("before") or {}
+        after = (self.result.get("diff") or {}).get("after") or {}
+        self._mark_changed_by_mac(endpoint_name, name)
+        self.result["diff"] = self._build_diff(
+            before={
+                **before,
+                "primary_mac_address": serialized.get("primary_mac_address"),
+            },
+            after={**after, "primary_mac_address": mac.id, "mac_address": mac_address},
+        )
+
+    def _mark_changed_by_mac(self, endpoint_name, name):
+        """Flag the interface as changed by the MAC update, without clobbering a more
+        specific message already set by ``_ensure_object_exists`` (e.g. ``... created``).
+        """
+        self.result["changed"] = True
+        prior = self.result.get("msg", "")
+        if not prior or "already exists" in prior:
+            self.result["msg"] = "%s %s updated" % (endpoint_name, name)
+
+    def _find_or_create_mac_address(
+        self, mac_address, assigned_object_type, assigned_object_id
+    ):
+        """Return the MACAddress for ``mac_address`` assigned to the object, creating it
+        if needed. MAC addresses are not unique in NetBox v4.2+, so match on both the
+        address and the assigned object to stay idempotent.
+        """
+        for candidate in self.nb.dcim.mac_addresses.filter(mac_address=mac_address):
+            if (
+                str(candidate.assigned_object_type) == assigned_object_type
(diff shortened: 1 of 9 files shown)


plugins/module_utils/netbox_utils.py
```diff
@@ -837,6 +837,82 @@ def _version_check_greater(
...
+        serialized = nb_object if isinstance(nb_object, dict) else nb_object.serialize()
+
+        # The interface's read-only `mac_address` already reflects its primary MAC.
+        if serialized.get("mac_address") == mac_address:
+            return
+
+        if self.check_mode:
+            self.result["changed"] = True
+            self.result["msg"] = "%s %s updated" % (endpoint_name, name)
+            return
```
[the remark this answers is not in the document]
ldrozdz93: [claude] Applied. Added a small `_mark_changed_by_mac` helper that only overwrites the message with "… updated" when the prior message is the no-change case ("… already exists") or unset, so a "… created" message from `_ensure_object_exists` is preserved. Used it in both the check-mode and real-write branches for consistency.

plugins/module_utils/netbox_utils.py
```diff
@@ -837,6 +837,85 @@ def _version_check_greater(
...
+        try:
+            return self.nb.dcim.mac_addresses.create(
+                {
+                    "mac_address": mac_address,
+                    "assigned_object_type": assigned_object_type,
+                    "assigned_object_id": assigned_object_id,
+                }
+            )
+        except pynetbox.RequestError as e:
+            self._handle_errors(msg=e.error)
```
unlisted: [claude]
`_find_or_create_mac_address` returns `None` if `mac_addresses.create` raises a `RequestError` whose `e.error` is falsy: `_handle_errors` only calls `fail_json` when `msg` is truthy, so it falls through, returns `None`, and the caller then dereferences `mac.id` -> `AttributeError` with a confusing traceback instead of a clean failure. This mirrors the existing pattern in `_create_netbox_object`, so it is not new to this PR and I would not block on it, just noting the one new dereference (`mac.id`) that assumes a non-`None` return.

unlisted (dismissed):
[claude]

Approving. Two findings that may be worth addressing (neither blocks):

**Scope: `netbox_vm_interface` has the identical bug, unaddressed.** VM interfaces also expose `mac_address` + `primary_mac_address` (`plugins/modules/netbox_vm_interface.py:51,216`) and moved to the `MACAddress` model in v4.2 the same way, so they will still report a phantom change every run. Since the helper already lives on the shared base class, `netbox_virtualization.py`'s `run()` could call it with `assigned_object_type="virtualization.vminterface"`. Fine as a follow-up, just worth a line in the PR that the fix is DCIM-only.

Second finding is the inline comment on `_find_or_create_mac_address`.

plugins/module_utils/netbox_utils.py
ldrozdz93: [claude] Applied in 0fcf6b9: the create call now falls back to a non-empty message (`e.error or "Failed to create MAC address <mac>"`), so `_handle_errors` always reaches `fail_json` and the `mac.id` dereference can no longer see `None`. Left `_create_netbox_object` untouched as a pre-existing repo-wide pattern, per the note.

ldrozdz93:
> Scope: netbox_vm_interface has the identical bug, unaddressed. VM interfaces also expose mac_address + primary_mac_address (plugins/modules/netbox_vm_interface.py:51,216) and moved to the MACAddress model in v4.2 the same way, so they will still report a phantom change every run. Since the helper already lives on the shared base class, netbox_virtualization.py's run() could call it with assigned_object_type="virtualization.vminterface". Fine as a follow-up, just worth a line in the PR that the fix is DCIM-only.

Issue created to track it: https://github.com/netbox-community/ansible_modules/issues/1586

ldrozdz93:
The approval was dismissed by just a minor commit about error formatting, so I'm bypassing to merge.

PR #1591, opened 2026-07-21
fix: revert slug generation to pre-3.23.0 behavior

### LINKED ISSUE #1563: [Bug]: Ansible Release Management WG: Potential semantic versioning issue in 3.23.0

3.23.0

### Ansible version

```bash
<!--- Paste verbatim output from "ansible --version" between quotes -->
```

### NetBox version

n/a

### Python version

3.8

### Steps to Reproduce

Hi :wave: from the Ansible Release Management WG. While releasing ansible 13.7.0, we noticed that `netbox.netbox` 3.23.0 includes what appears to be a breaking change listed in the changelog:

> - This changes the behavior for generating slugs. If you were relying on the previous behavior, you need to be aware that this behavior has changed.

In https://github.com/ansible-community/ansible-build-data/pull/680, we pinned the collection version included in the Ansible package to `<3.23.0` since we do not allow breaking changes in Ansible stable releases. See https://docs.ansible.com/projects/ansible/latest/community/collection_contributors/collection_requirements.html#versioning-and-deprecation for more details.

Can you please clarify whether this is indeed a breaking change? If so, can you please release `3.23.1` that reverts this breaking change and save it for a new semver major release of `netbox.netbox` (4.0.0)?

Thank you!

### Expected Behavior

n/a

### Observed Behavior

n/a

unlisted:
## Summary

- Revert `_to_slug()` to the pre-3.23.0 behavior: runs of separator characters
  (`-`, `.`, whitespace) collapse to a single hyphen, and leading/trailing
  separators are preserved as one hyphen. The function body is now
  byte-identical to `v3.22.0`. This undoes the `_to_slug()` transform
  introduced in #1499 (commit 5f6de01).
- The 3.23.0 change added a trim of leading/trailing whitespace and dots ahead
  of the collapse. Run-collapsing itself is not new: it is unchanged across all
  30 tags from v2.0.0 through v3.22.0. The trim altered auto-generated slugs, so
  playbooks that let the module derive a slug and rely on the previous value as
  a lookup key saw idempotency break. That shipped in a minor release, which the
  Ansible Release Management WG flagged in #1563.
- Keep the additive, non-breaking fixes from #1499: the `slug`-key early
  `continue` in `_normalize_data()` and the "slugify only if needed" guard in
  `netbox_secrets.py`. Only the `_to_slug()` transform is reverted.
- Extend the `slug` fixture from 6 cases to 12 so the tests actually cover the
  reverted behavior, adding the inputs that discriminate between the two
  implementations: repeated separators (`Test  Rack`, `Test - Rack`) and
  leading/trailing separators (`...a.b...`, `' leading'`, `'trailing '`,
  `' both '`). The original 6 cases contain only single interior separators, so
  they pass under either implementation and could not have caught this. Every
  expectation was generated by executing `v3.22.0`'s `_to_slug` rather than the
  current code, so the fixture asserts the restored release's behavior instead
  of snapshotting whatever this branch happens to do. No test code changes:
  `test_netbox_module.py` is identical to `devel`.

### A note for reviewers: why `5f6de01`'s diff is misleading

#1499 contains **two** commits touching this regex. `02f8839` removed the `+`
quantifier; `5f6de01` restored it *and* added the end-trim. Reading `5f6de01`'s
diff alone therefore suggests "before: one hyphen per separator character;
after: runs collapse" — but its parent is a transient intra-PR state that
appears in **zero release tags**. v3.22.0 already had the `+`.

The first revision of this PR reverted to that never-shipped baseline, which is
what @ldrozdz93 caught. If you are checking this revert, compare release tags
(`git show v3.22.0:` vs `git show v3.23.0:`), not `5f6de01` against its parent.

The upstream-aligned slug generation is not dropped, only deferred to a future
major release (4.0.0) with a `breaking_changes` changelog entry, per the WG
request in #1563. This PR intentionally does not close #1563, which stays open
to track the 4.0.0 re-land and the ansible-build-data un-pin. The extended
fixture is what that re-land will have to rewrite, which makes the behavior
change deliberate rather than silent.

Closes INT-480

## Test plan

- [x] `ansible-test units`: 246 pass
- [x] `_to_slug()` body diffed against `git show v3.22.0:` output, identical
- [x] `slug.json` copied unmodified onto a pristine `v3.22.0` worktree with
      `plugins/` untouched passes 12/12 there. This is the check that the
      fixture is a true statement about the release being restored, rather than
      a snapshot of this branch.
- [x] Mutation tested so the fixture is known to be able to fail. Dropping the
      `+` fails 3 cases; the full v3.23.0 implementation fails 4; a partial
      re-land trimming only whitespace fails 3; trimming only dots fails 3.
- [x] Changelog fragment re-validated as YAML after the wording correction;
      `ansible-test sanity --test changelog` runs in CI
- [x] Full CI green on `b026c53`: **76/76 checks**, including the 12-job
      `integration_testing` matrix (NetBox v4.1 through v4.5) and 14 sanity
      jobs. The pre-review local integration run validated the first,
      incorrect revert, which is why the CI matrix is the result that counts
      here.

### DIFF
--- a/tests/unit/module_utils/test_data/netbox_utils/slug.json
+++ b/tests/unit/module_utils/test_data/netbox_utils/slug.json
@@ -22,5 +22,17 @@
     {
         "non_slug": "test1.example.com",
         "expected": "test1-example-com"
+    },
+    {
+        "non_slug": "Test  Rack",
+        "expected": "test--rack"
+    },
+    {
+        "non_slug": "Test - Rack",
+        "expected": "test---rack"
+    },
+    {
+        "non_slug": "...a.b...",
+        "expected": "---a-b---"
     }
 ]
--- a/tests/unit/module_utils/test_data/netbox_utils/slug.json
+++ b/tests/unit/module_utils/test_data/netbox_utils/slug.json
@@ -25,14 +25,26 @@
     },
     {
         "non_slug": "Test  Rack",
-        "expected": "test--rack"
+        "expected": "test-rack"
     },
     {
         "non_slug": "Test - Rack",
-        "expected": "test---rack"
+        "expected": "test-rack"
     },
     {
         "non_slug": "...a.b...",
-        "expected": "---a-b---"
+        "expected": "-a-b-"
+    },
+    {
+        "non_slug": " leading",
+        "expected": "-leading"
+    },
+    {
+        "non_slug": "trailing ",
+        "expected": "trailing-"
+    },
+    {
+        "non_slug": " both ",
+        "expected": "-both-"
     }
 ]
--- a/changelogs/fragments/int-480-revert-slug-generation-semver-break.yml
+++ b/changelogs/fragments/int-480-revert-slug-generation-semver-break.yml
@@ -2,10 +2,9 @@
 bugfixes:
   - >-
     Reverted the slug-generation change introduced in 3.23.0, restoring the
-    pre-3.23.0 behavior (repeated separators are no longer collapsed and leading
-    or trailing separators are no longer trimmed). The 3.23.0 change altered
-    auto-generated slugs, breaking idempotency for playbooks that rely on the
-    derived slug as a lookup key, and it shipped in a minor release contrary to
-    semantic versioning. The upstream-aligned slug generation will return in a
-    future major release
+    pre-3.23.0 behavior (leading and trailing separators are no longer
+    trimmed). The 3.23.0 change altered auto-generated slugs, breaking
+    idempotency for playbooks that rely on the derived slug as a lookup key,
+    and it shipped in a minor release contrary to semantic versioning. The
+    upstream-aligned slug generation will return in a future major release
     (https://github.com/netbox-community/ansible_modules/issues/1563).
--- a/changelogs/fragments/int-480-revert-slug-generation-semver-break.yml
+++ b/changelogs/fragments/int-480-revert-slug-generation-semver-break.yml
@@ -1,0 +1,11 @@
+---
+bugfixes:
+  - >-
+    Reverted the slug-generation change introduced in 3.23.0, restoring the
+    pre-3.23.0 behavior (repeated separators are no longer collapsed and leading
+    or trailing separators are no longer trimmed). The 3.23.0 change altered
+    auto-generated slugs, breaking idempotency for playbooks that rely on the
+    derived slug as a lookup key, and it shipped in a minor release contrary to
+    semantic versioning. The upstream-aligned slug generation will return in a
+    future major release
+    (https://github.com/netbox-community/ansible_modules/issues/1563).
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -1553,7 +1553,7 @@ class NetboxModule(object):
             return value
 
         removed_chars = re.sub(r"[^\-\.\w\s]", "", value)
-        convert_chars = re.sub(r"[\-\.\s]", "-", removed_chars)
+        convert_chars = re.sub(r"[\-\.\s]+", "-", removed_chars)
         return convert_chars.strip().lower()
 
     def _normalize_data(self, data):
--- a/plugins/module_utils/netbox_utils.py
+++ b/plugins/module_utils/netbox_utils.py
@@ -1552,10 +1552,9 @@ class NetboxModule(object):
         elif isinstance(value, int):
             return value
 
-        value = re.sub(r"[^\-.\w\s]", "", value)
-        value = re.sub(r"^[\s.]+|[\s.]+$", "", value)
-        value = re.sub(r"[-.\s]+", "-", value)
-        return value.strip().lower()
+        removed_chars = re.sub(r"[^\-\.\w\s]", "", value)
+        convert_chars = re.sub(r"[\-\.\s]", "-", removed_chars)
+        return convert_chars.strip().lower()
 
     def _normalize_data(self, data):
         """


tests/unit/module_utils/test_data/netbox_utils/slug.json
```diff
@@ -22,5 +22,17 @@
     {
         "non_slug": "test1.example.com",
         "expected": "test1-example-com"
+    },
+    {
+        "non_slug": "Test  Rack",
```
ldrozdz93, line 27: These three cases pass this PR, but don't pass pre-3.23.0, i.e. v3.22.0.


| `non_slug` | in this PR | should be (v3.22.0) |
|---|---|---|
| `Test  Rack` | `test--rack` | `test-rack` |
| `Test - Rack` | `test---rack` | `test-rack` |
| `...a.b...` | `---a-b---` | `-a-b-` |

## Proof:

This fixture file, copied **unmodified** onto a `v3.22.0` worktree and run against the existing, unchanged test:

```console
$ git worktree add --detach /tmp/v322 v3.22.0
$ cp tests/unit/module_utils/test_data/netbox_utils/slug.json \
     /tmp/v322/tests/unit/module_utils/test_data/netbox_utils/slug.json
$ cd /tmp/v322 && pytest tests/unit/module_utils/netbox_utils/test_netbox_module.py::test_to_slug_returns_valid_slug -q

FAILED ...[Test  Rack-test--rack]   - AssertionError: assert 'test-rack' == 'test--rack'
FAILED ...[Test - Rack-test---rack] - AssertionError: assert 'test-rack' == 'test---rack'
FAILED ...[...a.b...----a-b---]     - AssertionError: assert '-a-b-'    == '---a-b---'
3 failed, 6 passed in 0.15s
```

**Worth noting:** the corrected fixture passing 9/9 against `v3.22.0` also confirms that restoring the `+` quantifier in `_to_slug` is the *complete* fix — no other change is needed.

ldrozdz93 (changes requested):
[claude] local-pr-review: 1 finding(s)

tests/unit/module_utils/test_data/netbox_utils/slug.json
unlisted: Good call, and thanks for doing the v3.22.0 proof rather than just flagging it. You're right on all three cases.

I checked the full tag history: the `+` quantifier is present in every release from v2.0.0 through v3.22.0, 30 tags with a byte-identical regex. #1499 is the only deviation. So dropping it produced a third behavior matching neither v3.22.0 nor v3.23.0.

Worse, it was more disruptive than the regression it was meant to fix. Over 15 sample inputs it diverged from v3.22.0 on 7, where the 3.23.0 change diverges on only 3. Realistic names were hit (`Rack 1 - Row A` became `rack-1---row-a`, `Site: NYC  DC1` became `site-nyc--dc1`), and since hyphens are legal slug characters nothing would have rejected them. It would have silently broken idempotency for a wider set of playbooks than the ones this PR set out to rescue.

Root cause on my side, for the record: I retyped the old function instead of extracting it from the tag, which lost the `+`. Then I derived the three new fixture expectations by running the new code instead of v3.22.0. That turned the only test capable of catching the bug into a test that pinned it in place, and the changelog and PR description were written to match, so it read as intentional. The pre-existing 6 fixture cases all contain single separators only, so the suite had zero discriminating power on the exact axis I was changing.

Fixed in 80db229:

- Restored the `+` in `_to_slug`. The function body is now byte-identical to `git show v3.22.0:plugins/module_utils/netbox_utils.py`.
- Corrected the three fixture expectations to the real v3.22.0 values.
- Corrected the changelog fragment, which claimed run collapsing was reverted. Only the leading/trailing trim was new in 3.23.0.
- Added `test_to_slug_matches_legacy_contract` with a `slug_legacy_contract.json` fixture: 10 cases pinning both axes #1499 touched, including the leading and trailing cases (`' leading'`, `'trailing '`, `' both '`) that the general fixture never exercised. Every expectation was generated by executing v3.22.0's `_to_slug` from the release source, not by running the new code.

I used your method to validate it. The new test and both fixtures, copied unmodified onto a pristine v3.22.0 worktree with `plugins/` untouched, pass 19/19 there. The new test deliberately uses only `load_relative_test_data` and `mock_netbox_module`, both of which exist in v3.22.0, so it stays portable to that checkout for future re-validation.

I also mutation tested the guard to confirm it can fail: dropping the `+` again fails 10 cases, and swapping in the v3.23.0 implementation fails 5. So the 4.0.0 re-land tracked in #1563 has to rewrite the contract deliberately alongside a `breaking_changes` fragment, which is what I should have had here from the start.

unlisted: Follow-up on the guard I described above, since I changed it in c63a647 and don't want you reviewing a stale description.

I claimed the new `test_to_slug_matches_legacy_contract` was a guard against this regression recurring. I checked that claim by mutation testing instead of assuming it, and it was mostly wrong: once the `slug.json` expectations were corrected, `slug.json` on its own already caught the bug you found (3 failures) and a full v3.23.0 re-land (1 failure). The separate contract test just failed louder on the same mutations, while duplicating 3 fixture rows and adding a second fixture pinning overlapping behavior.

It did close one real gap. `slug.json`'s only boundary case was `...a.b...`, which uses dots, so it was blind to a change that trims whitespace but not dots, and that is a plausible future edit given how often hand-entered NetBox names carry trailing spaces.

So in c63a647 I folded it in rather than keeping both: `slug_legacy_contract.json` and the second test are gone, `slug.json` goes from 6 cases to 12 (adding `' leading'`, `'trailing '`, `' both '`), and the provenance note moved onto the existing test's docstring. Net 36 lines lighter, one fixture instead of two.

Coverage is strictly better than either previous version. Mutation results against the single fixture: drop the `+`, 3 failures; full v3.23.0 implementation, 4; trim whitespace only, 3; trim dots only, 3. The old `slug.json` scored 0 on that third one.

Still validated your way: `slug.json` plus the test file, copied unmodified onto a pristine v3.22.0 worktree with `plugins/` untouched, pass 12/12 there.

ldrozdz93 (approved):

